PR #41, opened 2026-05-24
feat: add CUDA 13 support (jax 0.10, Python >=3.11)

antonioterpin:
## What
Adds a `cuda13` optional-dependencies extra so downstream projects (flowgames) can opt into CUDA 13 builds:

```toml
cuda13 = [
    "jax[cuda13]>=0.10.1",
]
```

## Why this needs Python >=3.11
The `cuda13` extra only exists in **jax >=0.10**, and jax 0.10 requires **Python >=3.11**. So this PR also:
- bumps `requires-python` to `>=3.11`
- pins the dev interpreter via `.python-version` (3.11)
- regenerates `uv.lock`

## ⚠️ Compatibility note
Because a single environment can't hold both the old base jax and a cuda13-capable jax, this floats **base jax 0.6.2 → 0.10.1** for the whole package (not just the GPU extra). jax 0.6 → 0.10 includes breaking API changes, so **the test suite should be run and any jax 0.10 breakage fixed** before merging to main.

This is pushed to `dev` so flowgames can pin to the dev branch in the meantime.


antonioterpin:
Closing: the CUDA-13 work (commits 1ab560d, a783f74, 152b752) is already merged into `dev`, so this branch is effectively dev itself. This PR (dev→main) would fast-forward all 92 accumulated dev commits into main, which is a much larger change than this PR describes and shouldn't ride in under the CUDA-13 banner. flowgames can pin to `dev` directly in the meantime. A proper dev→main integration can be opened later once the jax 0.6→0.10 suite is verified green.

PR #45, opened 2026-05-26
feat(optim): expose elementwise clip in the optimizer chain

fbanelli:
## Summary
- Register `optax.clip` as `"clip"` in `TRANSFORM_REGISTRY` so optimizer configs can request elementwise gradient clipping by name via `optimizer_config.chain`.

Ported from the private `fluids-estimation` dev branch (PR #153). Functional change only — the public optimizer docstring is left untouched.

## Test plan
- [x] `pytest tests/unit/training/test_optimizer_clip.py` (3 passed)

### DIFF
--- a/src/flowgym/training/optimizer.py
+++ b/src/flowgym/training/optimizer.py
@@ -19,6 +19,7 @@ OPTIMIZER_REGISTRY: dict[str, Any] = {
 }
 
 TRANSFORM_REGISTRY: dict[str, Any] = {
+    "clip": optax.clip,
     "clip_by_global_norm": optax.clip_by_global_norm,
     "add_decayed_weights": optax.add_decayed_weights,
     "scale": optax.scale,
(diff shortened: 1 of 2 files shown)


antonioterpin (approved):
## ✅ Approve

Clean, focused change that matches the description exactly: registers `optax.clip` as `"clip"` in `TRANSFORM_REGISTRY` (`src/flowgym/training/optimizer.py:22`), consumed by the existing chain builder.

- **Correctness:** `max_delta` matches `optax.clip`'s only kwarg; chain ordering (`optax.chain(*extra_txs, base_tx)`) runs the clip before the base SGD, so the test's `-lr * clip(grads)` expectation is right.
- **Tests:** `tests/unit/training/test_optimizer_clip.py` genuinely exercises the new behavior — registry identity, an end-to-end clipped chain, and a no-chain control. Good positive + negative coverage.
- **Conventions:** Google-style docstrings without type hints, explicit (deterministic) inputs, no lint silencing.

Non-blocking: the `build_optimizer_from_config` docstring example still only shows `clip_by_global_norm` — the PR intentionally leaves the public docstring untouched, so fine.

PR #46, opened 2026-05-26
feat(target_transforms): add log transform with safe default eps

fbanelli:
## Summary
- Add a `"log"` target transform that clamps negatives to zero and adds a small eps (default `1e-8`) before `jnp.log`, so a bare `name: log` config never injects `-inf` on exact-zero targets (EPE can be 0).
- Guard `build_target_transform_from_config` against non-dict configs with a descriptive `TypeError`.

Ported from the private `fluids-estimation` dev branch (commits 8e6e236, 6a89ed0). Public's existing docstrings are preserved.

## Test plan
- [x] `pytest tests/unit/training/test_target_transforms.py` (11 passed)

### DIFF
--- a/src/flowgym/training/target_transforms.py
+++ b/src/flowgym/training/target_transforms.py
@@ -82,6 +82,21 @@ def _scale(x: jax.Array, factor: float = 1.0) -> jax.Array:
     return x * factor
 
 
+def _log(x: jax.Array, eps: float = 1e-8) -> jax.Array:
+    """Log transform with epsilon and non-negative clamp.
+
+    Args:
+        x: Input array.
+        eps: Small positive constant added before log. Defaults to ``1e-8``
+            so a bare ``name: log`` config never injects ``-inf`` on exact
+            zeros (EPE targets/rewards can be 0).
+
+    Returns:
+        Log of (max(x, 0) + eps).
+    """
+    return jnp.log(jnp.maximum(x, 0.0) + eps)
+
+
 TransformFn = Callable[..., jax.Array]
 
 TRANSFORM_REGISTRY: dict[str, TransformFn] = {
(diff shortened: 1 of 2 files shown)


antonioterpin (commented):
## 💬 Review — log transform is correct, but the diff scope and base need attention

The `log` transform itself is well-implemented and tested. Two things to resolve before merge:

### 1. The diff contains PR #45's changes (undisclosed)
This branch is stacked on `feat/optim-elementwise-clip`, so `gh pr diff 46` against `dev` also includes the `optax.clip` registry addition (`src/flowgym/training/optimizer.py:22`) and the whole `tests/unit/training/test_optimizer_clip.py` (72 lines) — neither mentioned in this PR's summary. Please either:
- retarget this PR's base to `feat/optim-elementwise-clip` (#45) so the diff is just the log transform, or
- merge #45 first and rebase this onto `dev`,

so the diff and description match. (The body's '11 passed' count also predates the clip test — the file set is now 14.)

### 2. Log transform — verified correct
- Clamp + eps is right: `jnp.log(jnp.maximum(x, 0.0) + eps)` with default `eps=1e-8` keeps exact-zero targets finite (`log(1e-8)`) instead of `-inf`. (`src/flowgym/training/target_transforms.py:97`)
- Non-dict guard raises a descriptive `TypeError` already covered by the docstring `Raises:`. (`target_transforms.py:151-156`)
- Tests cover the explicit-eps + negative-clamp path, the zero-target finite-output default, and the non-dict guard.
- Note: the module is forward-only (no inverse), consistent with the existing `_log1p`/`_sqrt` siblings — so there's nothing to be inconsistent with, but worth being aware that `log` outputs are not invertible here.

Holding approval pending the scope/base cleanup in (1).

PR #47, opened 2026-05-26
feat(nn): convert RAFT32-PIV PyTorch checkpoint to Flax params

fbanelli:
## Summary
- Add `convert_raft_torch_state_dict` (`flowgym/nn/raft_torch_to_jax.py`) to map a RAFT32-PIV PyTorch state dict onto the Flax `RaftEstimatorModel` parameter tree, enabling reuse of pretrained PyTorch weights in the JAX estimator.
- Add `from __future__ import annotations` to `raft_piv_pytorch.py` (required by the converter's annotations).

Ported from the private `fluids-estimation` dev branch (PR #156, commits 0a226a7, 6ba2d43). Self-contained — no ArtOfPIV coupling.

## Test plan
- [x] `pytest tests/unit/nn/test_raft_torch_to_jax.py` with `other_methods` extra (torch) — 5 passed

### DIFF
--- a/src/flowgym/flow/raft/raft_piv_pytorch.py
+++ b/src/flowgym/flow/raft/raft_piv_pytorch.py
@@ -1,5 +1,7 @@
 """RAFT_torch class."""
 
+from __future__ import annotations
+
 import jax
 import jax.numpy as jnp
 import numpy as np
--- a/src/flowgym/nn/raft_torch_to_jax.py
+++ b/src/flowgym/nn/raft_torch_to_jax.py
@@ -1,0 +1,226 @@
+"""Convert a PyTorch RAFT32-PIV checkpoint into Flax/JAX parameters.
+
+The PyTorch model is defined in ``flowgym.nn.raft_torch_nn`` and the
+Flax counterpart in ``flowgym.nn.raft_model``. Both implement the same
+RAFT architecture but with different parameter naming. This module
+builds an explicit mapping between PyTorch state-dict keys and the
+nested Flax parameter tree produced by ``RaftEstimatorModel.init``.
+"""
+
+from __future__ import annotations
+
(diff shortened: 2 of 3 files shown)


antonioterpin (approved):
## ✅ Approve

Reviewed the converter, both source models (PyTorch `flowNetsRAFT`/`submodules_*` and the Flax `RaftEstimatorModel`/`blocks.py`), and the tests. No correctness issues found.

- **Conv transposition correct:** PyTorch OIHW → Flax HWIO via `_CONV_PERM = (2, 3, 1, 0)`, confirmed by the numerical parity test.
- **Module-tree mapping correct:** `EncoderBlock_0`=fnet / `EncoderBlock_1`=cnet matches Flax creation order; the 6 residual blocks, motion encoder, SepConvGRU, FlowHead, and mask head (skipping the ReLU at index 1) all line up. No norm params to map — both sides use affine-free instance norm, so the absence is correct, not an omission.
- **Tests are strong:** the two `slow` parity tests do real numerical-equivalence forward passes across all 12 iterations (not just shapes) and pass with torch installed; the numpy-only tests cover tree/shape and byte-for-byte leaf values. The 2e-2 tolerance is justified by the float16 inner conv.
- **Conventions:** Google-style docstrings without type hints, `from __future__ import annotations` present, lint/pyright clean.

**Minor (non-blocking):** the `from __future__ import annotations` added to `src/flowgym/flow/raft/raft_piv_pytorch.py:3` appears unnecessary — that file is in `flow/raft/`, unrelated to the converter in `nn/`, and type-checks cleanly without it. Harmless, but consider dropping it or correcting the 'required by the converter' justification, since it slightly contradicts the self-contained framing.

antonioterpin (approved):

Issue #44, opened 2026-05-24
bug: CacheManager.lookup does not prefer the newest part for duplicate keys


antonioterpin:
`test_lookup_disk_prefers_newest_part_for_duplicate_keys` fails on `dev`: when the same key is written across multiple flushed parts, `CacheManager.lookup` returns the value from an older part instead of the newest.

```
np.testing.assert_allclose(payload["values"], [[2.0]])
ACTUAL:  array([[1.]])   # oldest part
DESIRED: array([[2.]])   # newest part
```

Pre-existing (not introduced by any open PR; verified byte-identical to the test on `dev`). Surfaced when CI was enabled. The test is marked `xfail` referencing this issue until the dedup-by-newest-part logic in `src/flowgym/training/caching.py` is fixed.

PR #48, opened 2026-05-26
fix(caching): populate all spec columns on index hits + deterministic dups

fbanelli:
## Summary
- **Correctness fix**: in the index-lookup path, the per-column loop flipped `hit_mask` while writing the *first* spec field, so every subsequent field was skipped and read back as zeros. Now all columns are materialised once and every field is written per row before flipping the mask.
- **Determinism**: warm-start sorts parquet files newest-first; the `"index"` warm-start keeps the newest file per duplicate key; the pending-buffer lookup prefers the latest value for a key.

Ported rename-free from the private `fluids-estimation` dev `caching.py` (the private diff also carried an `estimator`→`model` rename, which is intentionally excluded here and will land as its own PR).

## Test plan
- [x] New regression test `test_index_lookup_populates_all_spec_columns` (fails on old code, passes now)
- [x] `pytest tests/unit/training/caching tests/integration/caching` (48 passed, 3 skipped, 1 xfailed)

### DIFF
--- a/src/flowgym/training/caching.py
+++ b/src/flowgym/training/caching.py
@@ -157,6 +157,14 @@ class CacheManager:
         if not files:
             return
 
+        # Deterministic duplicate-key policy across parts:
+        # prefer values from newer parquet files.
+        files = sorted(
+            files,
+            key=lambda p: (p.stat().st_mtime_ns, str(p)),
+            reverse=True,
+        )
+
         num_files = len(files)
         logger.info(
             f"Warm-starting cache ({self.warm_start}) from {num_files} files..."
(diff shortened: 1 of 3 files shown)


antonioterpin (commented):
## 💬 Review — core fix is correct; the determinism half is under-tested

The primary correctness fix is real and correctly implemented. I confirmed the bug by reverse-running the new regression test against the `dev` base: the old `_lookup_index` flipped `hit_mask[i]` after writing only the *first* spec field, so later columns read back as zeros (`scores: [0. 0. 0.]`). The fix in `src/flowgym/training/caching.py:307-326` materialises all columns once and flips the mask only after every field is written, with consistent row/column ordering — verified, no off-by-one.

The newest-first file sort (`caching.py:160-166`), per-key index dedup (`:198-209`), and pending-buffer latest-wins (`:228-239`) are all correct.

### Please address before merge

- **Determinism is claimed but ships untested.** The determinism work is exactly issue #44, but `tests/unit/training/caching/test_caching.py:158-192` (`test_lookup_disk_prefers_newest_part_for_duplicate_keys`) is left as `@pytest.mark.xfail`. The PR neither de-xfails it nor adds a replacement, so the newest-wins behavior has no passing regression test.
- **That xfail test is itself flaky and wouldn't reliably catch a regression.** It does `files = sorted(glob("*.parquet"))` (sort by random uuid name) then assigns mtimes by `files[0]`/`files[1]`, assuming name order == write order. It doesn't — running its logic 20× gave the wrong file 12/20 times. The implementation is correct; the test is wrong. Please fix it to assign mtimes by actual write order and flip it from `xfail` to a normal passing test (this PR resolves #44).

### Minor / optional
- The new regression test (`test_generic_caching.py:89-122`) only triggers the zeros bug because the scalar `epe` is declared before the vector `scores`. A stronger version would also assert correctness with the vector field first, so it can't silently stop exercising the bug if spec ordering changes.

Holding approval pending a deterministic, passing test for the newest-wins behavior.

antonioterpin (approved):

PR #49, opened 2026-05-26
refactor(make): rename estimator API to model (compile/save/load_model)

fbanelli:
## Summary
Renames the public `make.py` API to match the private `fluids-estimation` dev repo, so subsequent cross-repo ports (checkpointing, experiments) apply cleanly on a matching API surface:

- `compile_estimator` → `compile_model`
- `save_estimator` → `save_model`
- `load_estimator` → `load_model`
- their `estimator` parameter → `model`, and `estimator_name` → `model_name`

All call sites and the `save_model` test mock are updated. **Pure rename, no behavioural change** (132 insertions / 132 deletions).

Intentionally left unchanged (matching private): the `Estimator` class, `*EstimatorTrainableState`, `make_estimator`, the `estimators/` config dir, and `estimator:` config keys.

## Test plan
- [x] Collection clean: 1004 tests collected, no import errors
- [x] `pytest tests/unit/training/test_checkpointing.py` (5 passed) — exercises `save_model`/`load_model`
- [x] `pytest tests/unit/training/test_sampler_checkpointing.py tests/unit/training/test_partial_checkpointing.py` (5 passed)
- [x] `pytest tests/unit/flow/test_postprocess_config.py tests/unit/common/test_preprocess_config.py` — exercise `compile_model` (passed; CPU-heavy)

PR #50, opened 2026-05-26
feat(checkpoint): long-lived Checkpointer + W&B upload + wandb-resume

fbanelli:
## Summary
Ports the checkpointing stack from the private `fluids-estimation` dev (PRs #157–161) onto the renamed make API:

- **`flowgym/checkpointing.py`** — `Checkpointer` context manager with a single long-lived orbax `CheckpointManager`, native best-metric tracking, and `best`/`every`/`never` W&B upload modes (`save_periodic` / `on_validation` / `save_final`).
- **`flowgym/checkpoint_source.py`** — resolve `wandb://[entity/]project/name[:alias][/subpath]` artifact URIs to a local path for resume.
- **`flowgym/_synthpix_compat.py`** — scrub synthpix's NaN-shape restore placeholder that orbax 0.11+ rejects; installed once at `make` import.
- **`make.py`** — `build_save_args` + wandb-aware `load_from` resolution.
- **`train.py` / `train_supervised.py` / `main.py`** — route all saves through `Checkpointer`. Trainer-function params keep their existing `estimator` naming (the trainer-param rename was intentionally kept out of scope; only the make API was renamed in the stacked PR).
- **deps** — pin synthpix to the git rev the NaN-scrub targets; `uv.lock` regenerated.

> ⚠️ **Stacked on #49** (the estimator→model make API rename). Base will retarget to `dev` once #49 merges.

## Test plan
- [x] New-module unit tests: `test_checkpointer.py`, `test_checkpoint_source.py`, `test_synthpix_compat.py` — 70 passed
- [x] All checkpoint unit tests (incl. `test_checkpointing`, `test_partial`, `test_sampler`) — 80 passed
- [x] Integration: `test_full_integration.py`, `test_integrated_checkpointing.py`, `test_trainable_state_integration.py` — 19 passed (integration tests patch `Checkpointer` via an in-process fake)
- [x] `pre-commit` (ruff, ruff-format, pydoclint, basedpyright) passes on all changed files

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -46,7 +46,7 @@ exclude = ["tests*", "docs*"]
 # Dependency groups
 [dependency-groups]
 dev = [
-    "synthpix>=0.2.0",
+    "synthpix @ git+https://github.com/antonioterpin/synthpix.git@8cfbd7d",
     "pre-commit==4.0.1",
     "snowballstemmer",
     "pytest==7.4.4",
@@ -62,7 +62,7 @@ dev = [
 ]
 
 synthpix-cuda12 = [
-    "synthpix[cuda12]>=0.2.0",
+    "synthpix[cuda12] @ git+https://github.com/antonioterpin/synthpix.git@8cfbd7d",
 ]
 
 dev-cuda12 = [
--- a/src/flowgym/_synthpix_compat.py
+++ b/src/flowgym/_synthpix_compat.py
@@ -1,0 +1,114 @@
+"""Compatibility shim for synthpix sampler restore.
+
+synthpix's :class:`SyntheticImageSampler.restore_state` substitutes a
+``jax.ShapeDtypeStruct((np.nan,), jnp.uint8)`` template placeholder for
+the ``files_scheduler`` field whenever the underlying scheduler state
+was ``None`` at save time. The intent is "unknown dim — let orbax read
+the real shape from disk metadata".
+
+orbax 0.11+ rejects NaN-valued shape entries during deserialization with
+``TypeError: 'float' object cannot be interpreted as an integer`` at
+``orbax/checkpoint/_src/arrays/numpy_utils.py:53: slice(*x.indices(n))``.
+The intended "use the saved shape" semantics is already what orbax does
+for ``None`` template leaves, so we patch ``restore_state`` to replace
+NaN-shape placeholders with ``None``.
+
+The shim is idempotent and a no-op when synthpix is not installed.
+Tracked upstream at synthpix `src/synthpix/sampler/synthetic.py` (the
+``np.nan`` sentinel is still on `main`); remove this shim once a
+synthpix release drops the NaN placeholder.
+"""
+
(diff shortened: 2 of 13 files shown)


antonioterpin (commented):
## 💬 Review — strong core; one test regression to fix (and a stacking note)

The checkpointing core is well-designed, well-documented, and thoroughly unit-tested (70 new unit tests pass; lint + basedpyright clean on the new files). I reviewed the orbax lifecycle, best-metric tracking, W&B upload modes, and `wandb://` parsing in detail — all correct (see below). Two test issues to resolve.

### Stacking
This PR targets `refactor/estimator-to-model-rename` (#49), not `dev`. The diff is correctly scoped to checkpointing on top of the rename. It must not merge before #49, and should be retargeted to `dev` once #49 lands.

### Should fix
- **Test regression — 3 RL integration tests now fail without the OpenCV extra.** The rewritten `tests/integration/training/test_full_integration.py` dropped the `patch("train.log_flow_estimate")` mock the base-branch RL tests had (base: 3 sites; the rewritten `test_train_rl_full_integration` / `_replay_buffer_used` / `_replay_buffer_samples`: none). On `dev` these pass; here they fail with `ImportError: OpenCV required` from `flowgym.utils.log_flow_estimate`. They'll pass in CI where the extra is installed, but the rewrite removed the test's isolation from an optional dependency — please re-add the mock.

### Minor / test-fidelity
- **`_FakeCheckpointer.save_final` signature mismatch** (`test_full_integration.py:55`): defined as `save_final(self, state, step)`, but production calls `checkpointer.save_final(trainable_state, final_step, val_metrics=...)` (`src/train_supervised.py:491`). When a `val_sampler` is supplied the resulting `TypeError` is swallowed by the `try/except` at `train_supervised.py:504`, so the fake's final-save path is never actually exercised. Add `val_metrics=None` to the fake's signature. (Not a production bug — the real `save_final` is correct and unit-tested.)

### Verified correct (no action)
- **Orbax lifecycle:** `close()` calls `wait_until_finished()` then `close()` in `try/finally`, idempotent, invoked from `__exit__` via `ExitStack`; async-save barrier before W&B upload is present and tested. (`checkpointing.py:270, 352, 356-382`)
- **Best-metric:** strict `<`/`>` (ties aren't improvements), mode mapped from `higher_is_better`, `_seen_best` bootstrapped from orbax's persisted metrics on resume. Survives-resume and save-only-best regressions tested.
- **W&B modes:** `never`/`best`/`every` dispatch and `best`/`latest`/`final` aliasing correct and tested; upload failures caught and logged, not raised.
- **`wandb://` parsing:** alias/subpath boundaries, ambiguous-subpath rejection, absolute-path and `..`-traversal guards all implemented and tested.
- synthpix shim idempotent and no-op when synthpix absent; all `noqa`/`pyright: ignore` are inline and per-rule — no CLI lint silencing.

Holding approval pending the dropped `log_flow_estimate` mock (and ideally the fake `save_final` signature), plus the #49 dependency.

PR #51, opened 2026-05-27
feat(checkpoint): long-lived Checkpointer + W&B upload + wandb-resume

antonioterpin:
## Summary
Re-opens the work from the closed #50 (which targeted the now-deleted
`refactor/estimator-to-model-rename` branch), rebased cleanly onto `dev`
now that #49 is merged. Ports the checkpointing stack from the private
`fluids-estimation` dev (PRs #157–161):

- **`flowgym/checkpointing.py`** — `Checkpointer` context manager with a single long-lived orbax `CheckpointManager`, native best-metric tracking, and `best`/`every`/`never` W&B upload modes (`save_periodic` / `on_validation` / `save_final`).
- **`flowgym/checkpoint_source.py`** — resolve `wandb://[entity/]project/name[:alias][/subpath]` artifact URIs to a local path for resume.
- **`flowgym/_synthpix_compat.py`** — scrub synthpix's NaN-shape restore placeholder that orbax 0.11+ rejects; installed once at `make` import.
- **`make.py`** — `build_save_args` extraction + wandb-aware `load_from` resolution. Keeps the existing `estimator_config` parameter name (the stray `model_config` rename from the original #50 branch was dropped to match the merged #49 public API).
- **`train.py` / `train_supervised.py` / `main.py`** — route all saves through `Checkpointer`. Trainer-function params keep their existing `estimator` naming.
- **deps** — pin synthpix to the git rev the NaN-scrub targets; `uv.lock` regenerated.

## Test plan
- [x] New-module unit tests: `test_checkpointer.py`, `test_checkpoint_source.py`, `test_synthpix_compat.py`
- [x] Affected unit + integration tests: `test_full_integration.py` (RL + supervised) — 80 passed locally with the `other_methods` extra installed
- [x] Full suite green locally
- [x] `pre-commit` clean on all changed files (ruff, ruff-format, pydoclint, basedpyright); `basedpyright` passes project-wide with extras installed

### DIFF
--- a/src/flowgym/checkpointing.py
+++ b/src/flowgym/checkpointing.py
@@ -11,6 +11,7 @@ from typing import TYPE_CHECKING, Any, Literal
 
 import goggles as gg
 import orbax.checkpoint as ocp
+from orbax.checkpoint import checkpoint_managers as ocp_cm
 
 from flowgym.make import build_save_args
 
@@ -44,9 +45,11 @@ class CheckpointConfig:
             orbax's per-step ``metrics`` so ``best_step`` survives a
             process restart.
         higher_is_better: When True, larger metric values are better.
-        keep: Maximum number of on-disk checkpoint steps to retain
-            (orbax ``max_to_keep``). The best step is exempt from
-            eviction (orbax preserves it alongside the most recent N).
+        keep: Number of most-recent on-disk checkpoint steps to retain
+            for resume. The best step is preserved on top of these
+            latest N (see ``Checkpointer`` for the orbax retention
+            policy), so resuming from the newest checkpoint always
+            works even when the metric degrades over training.
         artifact_type: W&B artifact type label.
         artifact_name: W&B artifact name; defaults to
             ``f"{model.__class__.__name__}_checkpoint"``.
@@ -112,9 +115,19 @@ class Checkpointer:
 
     Owns **one** ``ocp.CheckpointManager`` for the lifetime of the
     training run, rooted at ``out_dir/checkpoints/``. Best-step
-    tracking is delegated entirely to orbax via ``CheckpointManager
-    Options.best_fn`` — orbax persists per-step ``metrics`` on disk so
+    tracking is delegated to orbax via ``CheckpointManagerOptions
+    .best_fn`` — orbax persists per-step ``metrics`` on disk so
     ``best_step`` survives a process restart without our own sidecar.
+
+    Retention combines two orbax preservation policies so the two
+    consumers never starve each other: ``LatestN(keep)`` keeps the
+    most recent ``keep`` steps for *resume*, and ``BestN(1)`` keeps the
+    single best-by-metric step for the ``best`` W&B alias. Driving
+    ``max_to_keep`` off ``best_fn`` alone (orbax's default when only
+    ``best_fn`` is set) would retain the best-N *by metric* and evict
+    the newest checkpoints whenever validation degrades — silently
+    breaking resume-from-latest — so the explicit composite policy is
+    required.
 
     Saves are async by default; the long-lived manager keeps an orbax
     background thread alive across calls. Callers must invoke
@@ -160,10 +173,29 @@ class Checkpointer:
             except (KeyError, TypeError, ValueError):
                 return float("-inf" if higher_is_better else "inf")
 
+        # Retain the most recent ``keep`` steps (for resume) *and* the
+        # single best step (for the ``best`` alias). ``best_fn`` alone
+        # would make orbax retain the best-N by metric and evict the
+        # newest checkpoints when validation degrades, breaking
+        # resume-from-latest; the explicit composite policy keeps both.
+        # ``reverse`` puts the better metric last so ``BestN`` keeps it:
+        # ascending sort (``reverse=False``) for higher-is-better,
+        # descending (``reverse=True``) for lower-is-better.
+        preservation_policy = ocp_cm.AnyPreservationPolicy(
+            [
+                ocp_cm.LatestN(n=self._cfg.keep),
+                ocp_cm.BestN(
+                    get_metric_fn=_best_fn,
+                    reverse=not higher_is_better,
+                    n=1,
+                    keep_checkpoints_without_metrics=False,
+                ),
+            ]
+        )
         options = ocp.CheckpointManagerOptions(
-            max_to_keep=self._cfg.keep,
             best_fn=_best_fn,
             best_mode="max" if higher_is_better else "min",
+            preservation_policy=preservation_policy,
             create=True,
             enable_async_checkpointing=True,
         )
@@ -233,26 +265,51 @@ class Checkpointer:
 
         Returns:
             The on-disk checkpoint directory path when a save happened,
-            ``None`` when validation was purely observational (no
-            extractable metric, or current mode does not save at
-            validation time).
+            ``None`` when validation was purely observational (current
+            mode does not save at this event, or ``"best"`` mode saw no
+            improvement / no extractable metric).
         """
         metric_value = self._extract_metric(val_metrics)
-        if metric_value is None:
-            return None
-        is_new_best = self._is_better(metric_value)
-        self._last_observed = {self._cfg.metric_key: metric_value}
+        mode = self._cfg.wandb_upload
+        is_new_best = metric_value is not None and self._is_better(metric_value)
+
+        # ``every`` snapshots the latest model after *every* validation,
+        # including one whose metric is missing or non-finite (e.g. a
+        # NaN-out near the end) — that run state is exactly what a
+        # post-mortem wants. ``best`` only acts on a genuine
+        # improvement, so a non-extractable metric is observational
+        # there and never saves.
+        will_save = mode == "every" or (mode == "best" and is_new_best)
+
+        prev_seen_best = self._seen_best
+        prev_has_unsaved_best = self._has_unsaved_best
+        # Attach only *this* event's metric to the save. A non-finite
+        # event saves with no metrics (in ``every`` mode) so it can
+        # never be mistaken for a best checkpoint.
+        current_metrics = (
+            {self._cfg.metric_key: metric_value}
+            if metric_value is not None
+            else None
+        )
+        if current_metrics is not None:
+            self._last_observed = current_metrics
         if is_new_best:
             self._seen_best = metric_value
             self._has_unsaved_best = True
-        mode = self._cfg.wandb_upload
 
-        will_save = mode == "every" or (mode == "best" and is_new_best)
         if not will_save:
             return None
 
-        path = self._save(state, step, self._last_observed)
+        path = self._save(state, step, current_metrics)
         if path is None:
+            # Orbax rejected the write (duplicate / backward step), so
+            # nothing landed on disk. Roll back the best trackers we
+            # advanced above; otherwise ``best_metric`` would report a
+            # value with no backing checkpoint and the stale
+            # ``_has_unsaved_best`` latch could trip a spurious later
+            # ``save_periodic`` write.
+            self._seen_best = prev_seen_best
+            self._has_unsaved_best = prev_has_unsaved_best
             return None
         self._has_unsaved_best = False
         # Orbax may have moved the best step now; re-confirm.
--- a/src/flowgym/checkpointing.py
+++ b/src/flowgym/checkpointing.py
@@ -1,0 +1,441 @@
+"""Checkpoint orchestration: orbax save, best-metric tracking, W&B upload."""
+
+from __future__ import annotations
+
+import math
+from collections.abc import Mapping
+from dataclasses import dataclass
+from pathlib import Path
+from types import TracebackType
+from typing import TYPE_CHECKING, Any, Literal
+
+import goggles as gg
+import orbax.checkpoint as ocp
+
+from flowgym.make import build_save_args
+
+if TYPE_CHECKING:
+    from synthpix.sampler import Sampler
+
+    from flowgym.common.base import Estimator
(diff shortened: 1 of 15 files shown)


src/flowgym/checkpointing.py
```diff
@@ -0,0 +1,441 @@
...
+
+        def _best_fn(metrics: Mapping[str, Any]) -> float:
+            try:
+                return float(metrics[metric_key])
+            except (KeyError, TypeError, ValueError):
+                return float("-inf" if higher_is_better else "inf")
+
+        options = ocp.CheckpointManagerOptions(
+            max_to_keep=self._cfg.keep,
+            best_fn=_best_fn,
```
antonioterpin, line 196: **Blocker — resume-from-latest is broken.** With `best_fn` set, orbax's `max_to_keep` retains the *best-N by metric*, not the most recent N. Verified empirically on orbax 0.11.31: saving steps 100→500 with worsening `mean_error` keeps only 100/200/300, `latest_step()` returns 300, and `restore(500)` raises `FileNotFoundError`. So whenever validation degrades over training (the normal overfitting case), the newest checkpoints get evicted and you can't resume from where you stopped. The docstring at L47-49 ("best step exempt … preserves the most recent N") is backwards.

Fix: keep latest-N for resume and track `best_step` separately (e.g. a `best` alias or separate retention), rather than driving `max_to_keep` off `best_fn`.

src/train_supervised.py
```diff
@@ -11,8 +11,8 @@
 from synthpix.sampler import Sampler
 
 from eval import evaluate_batches
+from flowgym.checkpointing import CheckpointConfig, Checkpointer
```
antonioterpin, line 14: **Blocker — breaks existing tests.** This module no longer exposes `save_model` (replaced by `Checkpointer`), but `tests/integration/caching/test_integration.py:50,165` and `tests/integration/caching/test_raft_integration.py:36` still `@patch("src.train_supervised.save_model")`. Those 3 tests now fail at patch setup with `AttributeError: <module 'src.train_supervised'> does not have the attribute 'save_model'`. Update them to patch `Checkpointer` (as `test_full_integration.py` does).

src/train.py
```diff
@@ -255,13 +273,14 @@ def train(
                 logger.info(f"Episode {episode_idx} - {k}: {avg_value}")
 
             if episode_idx % save_every == 0:
-                save_model(
-                    state=trainable_state,
-                    out_dir=out_dir,
-                    model=estimator,
-                    model_name=f"{estimator.__class__.__name__}-{episode_idx}",
-                    sampler=env_state[0],
+                saved_path = checkpointer.save_periodic(
```
antonioterpin, line 276: **`save_only_best=True` + RL = total checkpoint loss.** RL never calls `on_validation`/`save_final`, so `_has_unsaved_best` never flips. `CheckpointConfig.from_configs` derives `save_only_best` from the legacy `dataset_config['save_only_best']`, so an existing config makes every `save_periodic` here return `None` — the model is never persisted, not even at the end of the run. Either call `save_final` at the end of RL training, or ignore `save_only_best` when there's no validation loop.

tests/integration/training/test_full_integration.py
```diff
@@ -24,22 +27,56 @@
...
+        step_dir = ckpt_dir / str(step)
+        step_dir.mkdir(exist_ok=True)
+        self.save_calls.append((kind, step))
+        return str(step_dir)
+
+    def save_periodic(self, state, step):
+        return self._record_save("periodic", step)
+
+    def save_final(self, state, step):
+        return self._record_save("final", step)
```
antonioterpin, line 56: `_FakeCheckpointer.save_final(self, state, step)` is missing the `val_metrics` param the real `Checkpointer.save_final` has. `train_supervised.py:491` calls `save_final(..., val_metrics=...)`, which raises `TypeError` against this fake — swallowed by the broad `except Exception` at `train_supervised.py:504`. So these integration tests never actually exercise the final-save path; a regression there would go undetected. Add `val_metrics=None` to the fake's signature.

src/flowgym/checkpointing.py
```diff
@@ -0,0 +1,441 @@
...
+                ``evaluate_batches``.
+
+        Returns:
+            The on-disk checkpoint directory path when a save happened,
+            ``None`` when validation was purely observational (no
+            extractable metric, or current mode does not save at
+            validation time).
+        """
+        metric_value = self._extract_metric(val_metrics)
+        if metric_value is None:
```
antonioterpin: In `wandb_upload="every"` mode, a validation yielding a non-finite metric (`_extract_metric` → `None`) returns here before any save, so "every" silently skips the checkpoint/upload for that event — narrowing the documented "upload after every validation" contract. A NaN early in training, or a NaN-out near the end, produces no upload for those steps.

src/flowgym/checkpointing.py
```diff
@@ -0,0 +1,441 @@
...
+            return None
+        is_new_best = self._is_better(metric_value)
+        self._last_observed = {self._cfg.metric_key: metric_value}
+        if is_new_best:
```
antonioterpin, line 296: `_seen_best`/`_has_unsaved_best` are mutated *before* the save. If orbax rejects the save (`_save` returns `None` at L255 — duplicate/backward step), `_has_unsaved_best` is left `True` and `_seen_best` is advanced with no on-disk checkpoint backing it. `best_metric` then reports a value with no corresponding checkpoint, and the stale latch can trigger a spurious later `save_periodic` write. Reset the trackers when `_save` returns `None`.

antonioterpin (commented):
## Review: not mergeable as-is

Mergeable (no conflicts) but **BLOCKED** — branch protection requires review approval and no CI has run on this PR (no workflow has run in the repo since 2026-05-24; Actions doesn't appear to be firing on recent PRs).

### Confirmed blockers
1. **orbax retention evicts the newest checkpoints** → resume-from-latest broken (inline on `checkpointing.py`, verified empirically).
2. **3 out-of-diff caching tests fail** because `train_supervised.save_model` was removed but the patches weren't updated (inline on `train_supervised.py`).
3. **RL + `save_only_best` persists nothing** (inline on `train.py`).

### Lower severity (inline)
- Integration tests don't actually cover the final-save path (`_FakeCheckpointer.save_final` signature).
- `wandb_upload="every"` skips save on a non-finite metric.
- best-tracking desyncs when orbax rejects a save.
- `raft_torch` + `wandb://` load passes a directory to `torch.load`.

### Not this PR's fault
- `tests/unit/flow/test_openpiv_jax.py::test_fft_correlate_images` fails on a clean `dev` checkout too.
- `pre-commit` (pydoclint/basedpyright) only passes with the `other_methods` extra installed.

src/flowgym/make.py
```diff
@@ -475,54 +499,59 @@ def make_estimator(
...
+        # unchanged.
+        resolved_load_from = resolve_checkpoint_source(load_from)
         if estimator_config["estimator"] == "raft_torch":
             if torch is None:
-                raise ValueError("torch required for raft_torch estimator")
-            checkpoint = torch.load(load_from, map_location="cuda")
-            estimator_any = estimator  # type: Any
-            estimator_any.raft.load_state_dict(
+                raise ValueError("torch required for raft_torch model")
+            checkpoint = torch.load(resolved_load_from, map_location="cuda")
```
antonioterpin, line 550: For `estimator='raft_torch'` with a `wandb://` `load_from`, `resolve_checkpoint_source` (L534) returns the artifact's *directory*, then `torch.load(resolved_load_from, ...)` is handed a directory and raises `IsADirectoryError`. wandb resolution only resolves to a file when a `/subpath` is given. Require a subpath for torch checkpoints, or resolve to the checkpoint file.

src/flowgym/checkpointing.py
fbanelli: Fixed in 191de3b. Driving retention off `best_fn` alone was the bug — orbax then keeps best-N *by metric* and evicts the newest. The manager now gets an explicit `preservation_policy = AnyPreservationPolicy([LatestN(keep), BestN(1)])`, so the most-recent `keep` steps survive for resume **and** the single best step is preserved for the `best` alias. `best_fn`/`best_mode` stay set purely for `best_step()` tracking.

Empirically re-verified on orbax 0.11.39 (`reverse=not higher_is_better` so `BestN` keeps the right end): saving 100→500 with worsening `mean_error` now keeps `{100, 300, 400, 500}`, `latest_step()=500`, `best_step()=100`, and `restore(500)` succeeds. Docstrings on `keep` and the class were corrected, and `test_retention_keeps_latest_n_and_best` locks in the policy wiring.

src/train_supervised.py
fbanelli: Fixed in 191de3b. `tests/integration/caching/test_integration.py` (both tests) and `test_raft_integration.py` now `@patch("src.train_supervised.Checkpointer")` instead of the removed `save_model`. The mock was only ever used to suppress real I/O (never asserted on), so a patched `Checkpointer` (MagicMock supports the context-manager protocol) is a drop-in. All three pass.

src/train.py
fbanelli: Fixed in 191de3b. `train()` now calls `checkpointer.save_final(trainable_state, episode_idx)` after the episode loop (still inside the `ExitStack`). `save_final` always writes regardless of `save_only_best`, so RL now persists the final model even under the legacy `save_only_best` / `wandb_upload="best"` policies, and uploads it with the `final` alias when W&B is enabled. The RL integration test exercises this path now that the fake supports `save_final`.

tests/integration/training/test_full_integration.py
fbanelli: Fixed in 191de3b. `_FakeCheckpointer.save_final` now takes `val_metrics=None`, so `train_supervised.py:491`'s `save_final(..., val_metrics=...)` no longer raises `TypeError` (which the broad `except` was swallowing). The final-save path is now genuinely exercised by the supervised + RL integration tests.

src/flowgym/checkpointing.py
fbanelli: Fixed in 191de3b. `on_validation` no longer returns early on a non-extractable metric when `wandb_upload="every"`: that mode now snapshots/uploads after every validation, including a NaN-out, honoring the documented contract. The save carries **no** metrics in that case (`current_metrics=None`), so a non-finite step can never be ranked as best and uploads under `latest`, not `best`. `best`/`never` modes are unchanged (a non-extractable metric is purely observational there). Covered by `test_on_validation_every_saves_even_on_nonfinite_metric` and `test_on_validation_best_ignores_missing_or_nonfinite_metric`.

src/flowgym/checkpointing.py
fbanelli: Fixed in 191de3b. The best trackers are now snapshotted before mutation, and when `_save` returns `None` (orbax rejected a duplicate/backward step) both `_seen_best` and `_has_unsaved_best` are rolled back to their pre-call values. So `best_metric` never reports a value without a backing checkpoint, and the stale latch can no longer trip a spurious later `save_periodic`. Regression test: `test_on_validation_rolls_back_best_when_orbax_rejects_save`.

src/flowgym/make.py
fbanelli: Fixed in 191de3b. For `raft_torch`, if the resolved `load_from` is a directory (a `wandb://` URI without a `/subpath`), `make_estimator` now raises a clear `ValueError` pointing at the subpath grammar (`wandb://entity/project/name:alias/model.pt`) instead of handing a directory to `torch.load` and getting `IsADirectoryError`. The JAX/orbax path is unaffected since it consumes the directory.

fbanelli:
## Review addressed — pushed 191de3b

All confirmed blockers and the lower-severity items are fixed; replies are inline on each thread.

### Confirmed blockers
1. **Retention evicting newest** → explicit `AnyPreservationPolicy([LatestN(keep), BestN(1)])`; resume-from-latest restored (re-verified empirically on orbax 0.11.39).
2. **3 caching tests** → patch `Checkpointer` instead of the removed `save_model`.
3. **RL + `save_only_best` persisting nothing** → `train()` always calls `save_final` at the end.

### Lower severity
- `_FakeCheckpointer.save_final` gained `val_metrics=None`, so the final-save path is actually exercised.
- `wandb_upload="every"` now saves/uploads even on a non-finite metric (with no metrics attached, so it cannot rank as best).
- Best trackers roll back when orbax rejects a save.
- `raft_torch` + `wandb://` without a subpath now raises a clear error instead of feeding a directory to `torch.load`.

### Test / tooling status
- New + affected tests pass locally (`test_checkpointer.py`, caching + training integration). New regression tests: `test_retention_keeps_latest_n_and_best`, `test_on_validation_every_saves_even_on_nonfinite_metric`, `test_on_validation_rolls_back_best_when_orbax_rejects_save`, plus the `best`-mode split of the missing/non-finite test.
- `ruff` / `ruff-format` / `pydoclint` clean on the changed files.
- The remaining `basedpyright` warnings and the `test_fft_correlate_images` failure are the pre-existing, extras-/`dev`-related items already noted as not this PR's fault.

CI still has to be enabled on the repo for the required check to run — that part is outside the PR. Ready for another look.

fbanelli:
**Correction:** this follow-up should not have been pushed to the (already-merged) `feat/checkpointer-wandb-dev` branch — that commit (`8b16c2b`) is not on `dev`. The fix now lands properly via #52.

It addresses a regression in the always-on RL `save_final`: 4 RL replay tests drove the real `Checkpointer` with a mocked sampler/model/state (now patched, since they cover `ReplayBuffer` not checkpointing), and `save_final` now reuses an already-persisted step instead of dropping the `final` alias on a duplicate write.

PR #52, opened 2026-05-30
fix(checkpoint): RL replay tests + keep final alias (follow-up to #51)

fbanelli:
Follow-up to the now-merged #51. A regression in the always-on RL `save_final` (added in #51) was found while reviewing the private backport; this lands the fix on `dev` properly (it was mistakenly pushed onto the already-merged #51 branch first — that commit is not on `dev`).

## Changes
- **RL replay tests** (`test_train_rl_replay_buffer_used`, `test_train_rl_replay_buffer_samples`, `test_train_replay_initialization`, `test_train_replay_execution`) drove the *real* `Checkpointer` with a mocked sampler/model/state. The always-on `save_final` then forced a real orbax save that serialized a `MagicMock` (grain iterator / state `extras`) and failed. These tests cover `ReplayBuffer`, not checkpointing, so they now patch `train.Checkpointer` (as `test_train_rl_full_integration` already does).
- **`save_final`**: when the final step is already on disk (training ended on a `save_every` boundary that `save_periodic` just wrote), reuse the existing checkpoint instead of letting orbax reject a duplicate write — which dropped the `final` W&B alias and logged a spurious warning. The `final` artifact is now always tagged.
- **tests**: replace the duplicate-step test with a backward-unsaved-step case plus a regression for the reused-final-step path; add `all_steps` to the fake manager.

## Test plan
- [x] `tests/unit/training/test_checkpointer.py` + `tests/integration/training/{test_full_integration,test_train_replay_integration}.py` — 54 passed
- [x] ruff 0.14.1 / pydoclint / basedpyright clean on changed files

### DIFF
--- a/src/flowgym/checkpointing.py
+++ b/src/flowgym/checkpointing.py
@@ -384,8 +384,9 @@ class Checkpointer:
                 observed validation metrics.
 
         Returns:
-            The on-disk checkpoint path, or ``None`` when orbax rejected
-            the save (duplicate step / backward step).
+            The on-disk checkpoint path (the freshly written step, or an
+            already-persisted one at the same step), or ``None`` when
+            orbax rejected a genuinely backward save.
         """
         metrics: Mapping[str, float] | None
         if val_metrics is not None:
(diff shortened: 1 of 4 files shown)


antonioterpin (approved):
Reviewed the fix and tests. The `save_final` change correctly reuses an already-persisted final checkpoint (via `step in all_steps()`) instead of letting orbax reject a duplicate write, so the `final` W&B alias is always tagged; the reused path matches what `_save` returns and there's no best-metric-migration regression (that path is unchanged). Test changes are well-targeted: a renamed genuine-backward-rejection test, a new regression for the reuse path, and Checkpointer stubs on the ReplayBuffer-focused integration tests.

Verified locally on the PR head: ruff clean, basedpyright 0 errors on the changed source, and the full checkpointer + integration suites pass 54/54 (with the `other_methods` extra installed for OpenCV). LGTM.

PR #54, opened 2026-06-03
feat(open_piv): port sig2noise_ratio from openpiv to JAX

fbanelli:
Stacked on #53.

## What

Ports the signal-to-noise ratio — the third output of the reference `pyprocess.extended_search_area_piv` that the JAX pipeline previously dropped — following the same parity-test methodology as #53:

- `find_all_second_peaks`: `(2·width+1)²` exclusion box around the first peak via an `|idx − peak| ≤ width` mask + `-inf` argmax (equivalent to the reference's clamped slices), vectorized over arbitrary leading dims.
- `sig2noise_ratio`: `peak2peak` and `peak2mean`, with the reference's weak-peak (`< 1e-3`) and border-peak validity rules.
- `extended_search_area_piv(..., sig2noise_method=, width=)`: optional, `@overload`-typed — existing callers keep the plain-array API; when set, returns `(flow, s2n)` with `s2n` of shape `(B, n_rows, n_cols)`.

## One deliberate divergence from the reference

openpiv 0.25.4's `vectorized_sig2noise_ratio` builds its validity flag and then drops it: `peak2peak[flag is True] = 0` indexes with the constant `False`, which numpy treats as an empty boolean mask — a no-op. This port applies the flag as intended (matching the loop-based `pyprocess.sig2noise_ratio` semantics on the shared rules). The divergence is documented in the docstring and the test module, and parity is pinned accordingly: **exact on unflagged windows, zero on flagged ones**.

## Tests (in the #53 file structure)

- `test_openpiv_jax_numerical.py`: second-peak parity index-for-index (widths 1–3, pipeline + random + border-clipped maps); same-corr s2n parity across 2 methods × 3 widths × 3 window configs with reference-flag-aware comparison; crafted flag cases cross-checked against the loop-based reference; end-to-end pipeline s2n vs `pyprocess.extended_search_area_piv` (standard + extended search).
- `test_openpiv_jax_jit.py`: jit ≡ eager for `sig2noise_ratio` and the full pipeline with s2n (static method/geometry args).
- `test_openpiv_jax.py`: leading-dim vectorization, flow-only backward-compat return, invalid-method `ValueError`.

90/90 openpiv tests pass locally (CPU); ruff (pinned 0.14.1), pydoclint and basedpyright clean on the touched files.

### DIFF
--- a/src/flowgym/flow/open_piv/process.py
+++ b/src/flowgym/flow/open_piv/process.py
@@ -1,5 +1,7 @@
 """Module for OpenPIV processing in JAX."""
 
+from typing import overload
+
 import jax
 import jax.numpy as jnp
 from jax import lax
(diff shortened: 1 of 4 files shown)


antonioterpin:
Reopened against `dev` as #55 — the original base `feat-openpiv-jax` was deleted after #53 merged, so this work never reached `dev`. Superseded by #55.

PR #56, opened 2026-06-06
fix(validation): combine universal-median residuals with L2 norm (Westerweel eq. 2)

fbanelli:
## Problem

`universal_median_test` cites Westerweel & Scarano, *"Universal outlier detection for PIV data"* (Exp. Fluids 39(6), 2005), but combined the two per-component normalized residuals with a **per-component OR** (`any(r0 > r_threshold)`) instead of the **L2 norm** `sqrt(r0_u² + r0_v²)` specified in eq. 2 of that paper. Near the threshold these disagree, so the function flagged a different set of vectors than the method it cites.

This surfaced while auditing whether flowgym's validation methods — several of which are implementations of standard PIV algorithms — are tested against their reference. They are tested only against **in-file naive NumPy reimplementations**, and for this one the naive reference re-encoded the *same* OR combination, so the parity test validated the JAX vectorization against the implementation's own choice rather than against the reference algorithm. The bug was invisible to it by construction.

## Fix

- Combine the per-component normalized residuals with the **L2 norm**, matching eq. 2 of the paper.
- Make the naive reference **paper-faithful** (L2), with an opt-in `combine="or"` mode so a single reference can express both forms.
- Add **`test_universal_median_uses_l2_combination_not_per_component`** — a discriminating regression test that searches for a field where L2 and OR disagree and pins the implementation to L2. It fails on the previous behaviour (verified: reverting the one-line fix fails 9 tests including this one).
- Document the precise algorithm and the **two intentional divergences from `openpiv.validation.local_norm_median_val`**: the comparison median excludes the centre vector (the classic median-test convention and what the paper specifies; openpiv includes it), and borders are zero-padded (openpiv NaN-pads). Net effect: this implementation is now *fully* faithful to the paper, more so than openpiv's.

## Why not a direct openpiv-parity test

The repo pins `openpiv==0.25.1`, which doesn't even ship `local_norm_median_val` (only the plain `local_median_val`); the 0.25.4 version that does ship it diverges from the paper (centre-inclusion). So the authoritative reference here is the **paper's equation**, transcribed independently in the naive reference (per-pixel loops + `np.median` + explicit L2 — a different code path from the implementation's conv-patches + custom `median()` helper).

## Behaviour change

This changes which vectors `universal_median_test` flags near the threshold (now matching the cited method). Configs relying on the previous per-component behaviour may see slightly different masks; the new behaviour is the documented, paper-correct one.

## Verification

`tests/unit/common/test_data_validation.py`: 73 passed (33 GPU-timing tests skipped on CPU). ruff (pinned 0.14.1), pydoclint, basedpyright clean on the changed files.

### DIFF
--- a/src/flowgym/flow/postprocess/data_validation.py
+++ b/src/flowgym/flow/postprocess/data_validation.py
@@ -365,13 +365,34 @@ def universal_median_test(
     state: History | None = None,
     **kwargs: Any,
 ) -> tuple[jnp.ndarray, jnp.ndarray | None, History | None]:
-    """Universal outlier detection by median test.
+    """Universal outlier detection by the normalized median test.
 
-    See https://link.springer.com/article/10.1007/s00348-005-0016-6
+    Implements the normalized median test of Westerweel & Scarano,
+    "Universal outlier detection for PIV data", Experiments in Fluids 39(6),
+    2005 (https://link.springer.com/article/10.1007/s00348-005-0016-6). For
+    each vector the residual of every component is normalized by the local
+    median residual and the two normalized residuals are combined with the
+    L2 norm exactly as in eq. 2 of the paper:
+
+        r0* = sqrt( r0*_u^2 + r0*_v^2 ) > r_threshold   => outlier
+
+    where r0*_c = |U0_c - median(neighbours_c)| / (median(|U_c - median|) + eps)
+    and the neighbour statistics exclude the centre vector itself.
+
+    Divergences from ``openpiv.validation.local_norm_median_val`` (which
+    cites the same paper) are intentional and worth noting:
+
+    - the comparison median here excludes the centre vector (the classic
+      median-test convention and what the paper specifies), whereas openpiv
+      includes the centre in its ``um``; and
+    - borders are zero-padded (``conv_general_dilated_patches`` with
+      ``padding="SAME"``), whereas openpiv pads with NaN and uses
+      ``nanmedian`` so border windows use fewer real neighbours.
 
     Args:
         flow_field: Input array of shape (B, H, W, 2).
-        r_threshold: Threshold for the ratio of median to mean.
+        r_threshold: Threshold on the L2-combined normalized residual; a
+            vector is an outlier when the combined residual exceeds it.
         epsilon: Small value to avoid division by zero.
         radius: Radius for the local neighborhood
             (patch = (2*radius+1, 2*radius+1)).
@@ -420,9 +441,14 @@ def universal_median_test(
 
     valid = valid if valid is not None else jnp.ones((B, H, W), dtype=bool)
 
+    # Combine the per-component normalized residuals with the L2 norm, as in
+    # eq. 2 of Westerweel & Scarano (2005). A per-component OR would flag
+    # different vectors near the threshold.
+    r0_combined = jnp.sqrt(jnp.sum(jnp.square(r0), axis=-1))
+
     return (
         flow_field,
-        valid & ~jnp.any(r0 > r_threshold, axis=-1),
+        valid & ~(r0_combined > r_threshold),
         state,
     )
 
--- a/src/flowgym/flow/postprocess/data_validation.py
+++ b/src/flowgym/flow/postprocess/data_validation.py
@@ -443,8 +443,10 @@ def universal_median_test(
 
     # Combine the per-component normalized residuals with the L2 norm, as in
     # eq. 2 of Westerweel & Scarano (2005). A per-component OR would flag
-    # different vectors near the threshold.
-    r0_combined = jnp.sqrt(jnp.sum(jnp.square(r0), axis=-1))
+    # different vectors near the threshold. linalg.norm reads directly as the
+    # paper's eq. 2 and is more numerically robust than sqrt(sum(square(...)))
+    # for extreme residuals.
+    r0_combined = jnp.linalg.norm(r0, axis=-1)
 
     return (
         flow_field,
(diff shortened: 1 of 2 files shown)


tests/unit/common/test_data_validation.py
```diff
@@ -307,7 +321,10 @@ def _naive_median_test(
                 median = np.median(neigh, axis=0)
                 rm = np.median(np.abs(neigh - median), axis=0)
                 r0 = np.abs(patch[radius, radius, :] - median) / (rm + epsilon)
-                outlier[b, y, x] = np.any(r0 > r_threshold)
+                if combine == "l2":
+                    outlier[b, y, x] = np.sqrt(np.sum(r0**2)) > r_threshold
+                else:
+                    outlier[b, y, x] = bool(np.any(r0 > r_threshold))
```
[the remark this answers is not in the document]
fbanelli: Good catch — fixed. `_naive_median_test` now raises `ValueError` on any `combine` other than `"l2"`/`"or"`, so a typo can't silently change the reference semantics under the assertions.

fbanelli:
Addressed the Copilot review: the naive median reference now validates its `combine` argument and raises on unexpected values (commit pushed). All data-validation tests still pass; gates green.

tests/unit/common/test_data_validation.py
```diff
@@ -343,6 +362,46 @@ def test_universal_vs_naive(batch, height, width, radius, r_threshold):
...
+        ff = np.asarray(jrandom.normal(jrandom.PRNGKey(seed), (1, 9, 9, 2)))
+        out_l2 = _naive_median_test(
+            ff, r_threshold=r_threshold, radius=radius, combine="l2"
+        )
+        out_or = _naive_median_test(
+            ff, r_threshold=r_threshold, radius=radius, combine="or"
+        )
+        if np.any(out_l2 != out_or):
+            break
+    else:
```
antonioterpin: The discriminating test relies on a 50-seed brute-force search for an L2-vs-OR disagreement instead of a hand-constructed field. With 9x9 random normal inputs this almost always finds one, but it's brittle: a future change to JAX's RNG, the naive reference, or the default epsilon could push all 50 seeds into the no-disagreement region, and the test would then `pytest.fail` rather than catch a real regression. Consider constructing an explicit minimal field where one vector has r0_u ~= r0_v ~= 1.5 (so OR says valid, L2 ~= 2.12 flags), e.g. by placing a controlled perturbation at the centre of an otherwise-uniform 3x3 neighbourhood. That makes the assertion deterministic and the intent legible without the `for seed in range(50): ... else: pytest.fail` scaffolding.

src/flowgym/flow/postprocess/data_validation.py
```diff
@@ -420,9 +441,14 @@ def universal_median_test(
 
     valid = valid if valid is not None else jnp.ones((B, H, W), dtype=bool)
 
+    # Combine the per-component normalized residuals with the L2 norm, as in
```
antonioterpin, line 444: Minor: for the 2-component case `jnp.linalg.norm(r0, axis=-1)` (or `jnp.hypot(r0[..., 0], r0[..., 1])`) is equivalent here and slightly more numerically robust for very small/large residuals; `sqrt(sum(square(...)))` can lose precision or under/overflow when components are extreme. Residuals here are normalized so it's unlikely to bite, but `linalg.norm` also reads more directly as 'eq. 2 of the paper'. Optional.

antonioterpin (changes requested):

tests/unit/common/test_data_validation.py
fbanelli: Done in 3fad37c. Replaced the 50-seed brute-force search with an explicit minimal field: a 3x3 neighbourhood whose per-component median is 0 and MAD is 9.9, so at the default `epsilon=0.1` the divisor is exactly 10.0 and the centre vector `(15.0, 15.0)` has `r0_u == r0_v == 1.5`. Per-component OR (threshold 2.0) keeps it; the L2 combination `sqrt(1.5**2 + 1.5**2) ~= 2.12` flags it. The test now asserts the two references disagree exactly at that centre and that the implementation matches the L2 reference there — no RNG, no `for seed in range(50): ... else: pytest.fail` scaffolding.

src/flowgym/flow/postprocess/data_validation.py
fbanelli: Done in 3fad37c — switched to `jnp.linalg.norm(r0, axis=-1)`. It reads directly as eq. 2 of the paper and avoids the under/overflow risk of `sqrt(sum(square(...)))` for extreme residuals.

antonioterpin (approved):
**Second pass — LGTM (leaving pending for a final confirm).**

All three first-pass points are resolved and verified. The production change correctly implements Westerweel & Scarano eq. 2: `r0` is `(B,H,W,2)` and `jnp.linalg.norm(r0, axis=-1)` collapses the two per-component normalized residuals into a single scalar compared against `r_threshold` (replacing the per-component OR). The median-test convention is sound — neighbours exclude the centre via `jnp.delete(..., CENTER_FLAT)`, MAD/median over the 8 neighbours, divisor `rm + epsilon`, zero-padded borders.

I re-derived the discriminating test independently: 8 neighbours `±9.9` → median 0, MAD 9.9, divisor 10.0, centre `(15,15)` → `r0_u = r0_v = 1.5`; OR keeps it (max 1.5 < 2.0), L2 flags it (`√4.5 ≈ 2.12 > 2.0`). It genuinely discriminates L2 from OR, deterministically (no RNG). Naive-reference now raises on bad `combine`. Good.

PR #57, opened 2026-06-06
feat(open_piv): add centroid and parabolic subpixel methods

fbanelli:
## What

`subpixel_displacement` only implemented the gaussian (log-parabolic) fit. This adds the other two estimators from openpiv's `vectorized_correlation_to_displacements`, selectable via `subpixel_method=`:

- **`parabolic`** — plain 3-point parabolic fit `(cl − cr) / (2cl − 4c + 2cr)`.
- **`centroid`** — intensity-weighted centroid of the 3-point stencil. Note this yields an *absolute* position, so (unlike gaussian/parabolic) the peak index is not added back — matching the reference's separate `disp = shift − floor(N/2)` branch.

`extended_search_area_piv` gains a `subpixel_method` argument (default `"gaussian"`, so existing callers are unchanged; the `@overload`s are updated). It's a static Python string, closed over rather than vmapped. An unknown method raises `ValueError`, mirroring the reference.

## Tests (in the established file organization)

- `test_openpiv_jax_numerical.py`: the existing subpixel parity test is now parametrized over all three methods vs `pyprocess.vectorized_correlation_to_displacements`; added an end-to-end pipeline parity test per method, and an invalid-method `ValueError` test.
- `test_openpiv_jax_jit.py`: the subpixel jit test is parametrized over all three methods (static `subpixel_method` arg).
- `test_openpiv_jax.py`: backward-compat test that the default equals an explicit `gaussian` request.

All three methods match the reference to float32 round-off (~1e-6), with NaN/invalid-window positions agreeing exactly.

## Verification

106 openpiv tests pass locally (CPU; up from 90). ruff (pinned 0.14.1), pydoclint, basedpyright clean on the changed files.

Stacked conceptually after the sig2noise port (#55, merged); branches off current `dev`.

### DIFF
--- a/src/flowgym/flow/open_piv/process.py
+++ b/src/flowgym/flow/open_piv/process.py
@@ -19,6 +19,7 @@ def extended_search_area_piv(
     overlap: int,
     sig2noise_method: None = None,
     width: int = 2,
+    subpixel_method: str = "gaussian",
 ) -> jnp.ndarray: ...
 
 
@@ -31,6 +32,7 @@ def extended_search_area_piv(
     overlap: int,
     sig2noise_method: str,
     width: int = 2,
+    subpixel_method: str = "gaussian",
 ) -> tuple[jnp.ndarray, jnp.ndarray]: ...
 
 
@@ -42,6 +44,7 @@ def extended_search_area_piv(
     overlap: int,
     sig2noise_method: str | None = None,
     width: int = 2,
+    subpixel_method: str = "gaussian",
 ) -> jnp.ndarray | tuple[jnp.ndarray, jnp.ndarray]:
     """Batched PIV cross-correlation algorithm.
 
@@ -60,6 +63,9 @@ def extended_search_area_piv(
             output of the openpiv reference pipeline.
         width: Half-size of the exclusion box around the first correlation
             peak; only used when ``sig2noise_method == "peak2peak"``.
+        subpixel_method: Sub-pixel peak estimator passed to
+            :func:`subpixel_displacement`, one of ``"gaussian"``,
+            ``"parabolic"`` or ``"centroid"``.
 
     Returns:
         Displacement field of shape (batch_size, n_rows, n_cols, 2). If
@@ -130,9 +136,14 @@ def extended_search_area_piv(
     # reference's normalized_correlation=True path).
     corr = fft_correlate_images(aa, bb)
 
-    # Find peaks and compute displacements
+    # Find peaks and compute displacements. subpixel_method is static, so it
+    # is closed over rather than vmapped.
     peaks_i, peaks_j = find_all_first_peaks(corr)
-    disp_vx, disp_vy = jax.vmap(subpixel_displacement)(corr, peaks_i, peaks_j)
+    disp_vx, disp_vy = jax.vmap(
+        lambda c, pi, pj: subpixel_displacement(
+            c, pi, pj, subpixel_method=subpixel_method
+        )
+    )(corr, peaks_i, peaks_j)
 
     # Reshape displacements
     disp_vx = disp_vx.reshape(img1.shape[0], n_rows, n_cols)
@@ -367,23 +378,42 @@ def subpixel_displacement(
     peaks_j: jnp.ndarray,
     mask_width: int = 1,
     eps: float = 1e-7,
+    subpixel_method: str = "gaussian",
 ) -> tuple[jnp.ndarray, jnp.ndarray]:
     """Compute subpixel displacements from correlation maps.
+
+    Mirrors the three estimators of openpiv's
+    ``vectorized_correlation_to_displacements`` over the 3-point stencil
+    around each peak:
+
+    - ``"gaussian"``: log-parabolic fit, with a 3-point parabolic fallback on
+      any non-positive stencil value (the reference behaviour).
+    - ``"parabolic"``: plain 3-point parabolic fit.
+    - ``"centroid"``: intensity-weighted centroid of the stencil; unlike the
+      other two the fitted value is an absolute position rather than a
+      sub-pixel offset from the peak.
 
     Args:
         corr: Correlation maps (batch_size * n_windows, height, width).
         peaks_i: Peak indices in the i direction.
         peaks_j: Peak indices in the j direction.
         mask_width: Width of the mask for invalid peaks.
-        eps: Small constant added to the correlation map before the gaussian
-            fit, matching openpiv's vectorized_correlation_to_displacements.
-            It both prevents log(0) and keeps every stencil value strictly
+        eps: Small constant added to the correlation map before the fit,
+            matching openpiv's vectorized_correlation_to_displacements. It
+            both prevents log(0) and keeps every stencil value strictly
             positive, so the gaussian branch is always taken exactly as in
             the reference (no parabolic fallback on clipped zeros).
+        subpixel_method: Peak estimator, one of ``"gaussian"``,
+            ``"parabolic"`` or ``"centroid"``.
 
     Returns:
         Subpixel displacements (disp_vx, disp_vy).
+
+    Raises:
+        ValueError: If ``subpixel_method`` is not implemented.
     """
+    if subpixel_method not in ("gaussian", "parabolic", "centroid"):
+        raise ValueError(f"Method not implemented {subpixel_method}")
     if DEBUG:
         assert corr.ndim == 3, (
             "Correlation must be (batch_size * n_windows, height, width), "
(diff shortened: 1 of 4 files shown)


fbanelli:
Copilot generated no comments here. For consistency with the review on #58 (error-message quality), I also improved `subpixel_displacement`'s unknown-method `ValueError` to state the offending value and the valid options (`Unknown subpixel_method ...; expected one of 'gaussian', 'parabolic', 'centroid'.`). Tests/gates green.

src/flowgym/flow/open_piv/process.py
```diff
@@ -447,30 +480,45 @@ def subpixel_displacement(
...
+        # Intensity-weighted centroid: yields an absolute position, so the
+        # peak index is already folded in (no `+ safe_i` below).
+        fi, fj = safe_i.astype(corr.dtype), safe_j.astype(corr.dtype)
+        shift_i = ((fi - 1) * cl + fi * c + (fi + 1) * cr) / (cl + c + cr)
+        shift_j = ((fj - 1) * cd + fj * c + (fj + 1) * cu) / (cd + c + cu)
+        disp_vy = shift_i - jnp.floor(H / 2)
+        disp_vx = shift_j - jnp.floor(W / 2)
+    elif subpixel_method == "parabolic":
+        shift_i = (cl - cr) / (2 * cl - 4 * c + 2 * cr)
+        shift_j = (cd - cu) / (2 * cd - 4 * c + 2 * cu)
```
antonioterpin, line 512: The parabolic denominator `2*cl - 4*c + 2*cr` is unguarded here (and again in the gaussian fallback below). Unlike the log-parabolic branch — which uses `jnp.where(den != 0, nom/den, 0.0)` — a zero/near-zero denominator on a valid interior window will produce Inf/NaN that is not masked by the `invalid` border mask. openpiv has the same gap, so this is parity-preserving, but under JIT the divergence is silent and per-window. Worth either adding a `jnp.where(den != 0, ...)` guard (and matching it in the reference comparison) or asserting in the tests that a hand-crafted degenerate stencil reproduces the reference's Inf/NaN exactly, so we know future numerical drift won't desync the two implementations.

src/flowgym/flow/open_piv/process.py
```diff
@@ -447,30 +480,45 @@ def subpixel_displacement(
...
-    disp_vy = shift_i + safe_i - jnp.floor(H / 2)
-    disp_vx = shift_j + safe_j - jnp.floor(W / 2)
+    # 4) Estimate the sub-pixel peak with the requested method. The branch is
+    # on a static Python string, so only one path is traced.
+    if subpixel_method == "centroid":
+        # Intensity-weighted centroid: yields an absolute position, so the
+        # peak index is already folded in (no `+ safe_i` below).
+        fi, fj = safe_i.astype(corr.dtype), safe_j.astype(corr.dtype)
+        shift_i = ((fi - 1) * cl + fi * c + (fi + 1) * cr) / (cl + c + cr)
+        shift_j = ((fj - 1) * cd + fj * c + (fj + 1) * cu) / (cd + c + cu)
```
antonioterpin, line 502: The centroid divisor `cl + c + cr` (and `cd + c + cu`) is also unguarded and relies on `corr + eps` keeping every stencil value positive. After `normalized_correlation`, correlation entries can be substantially negative; if any of the three stencil values is less than `-eps`, the sum may approach zero or flip sign and the centroid silently produces a garbage absolute position. Consider either documenting this assumption (and how it can be violated post-normalization) or guarding the divisor with the same `jnp.where(... != 0, ...)` pattern used in the gaussian branch.

src/flowgym/flow/open_piv/process.py
```diff
@@ -447,30 +480,45 @@ def subpixel_displacement(
...
-    shift_j = jnp.where(inv, shift_j_fallback, shift_j_log)
-
-    disp_vy = shift_i + safe_i - jnp.floor(H / 2)
-    disp_vx = shift_j + safe_j - jnp.floor(W / 2)
+    # 4) Estimate the sub-pixel peak with the requested method. The branch is
+    # on a static Python string, so only one path is traced.
+    if subpixel_method == "centroid":
+        # Intensity-weighted centroid: yields an absolute position, so the
+        # peak index is already folded in (no `+ safe_i` below).
+        fi, fj = safe_i.astype(corr.dtype), safe_j.astype(corr.dtype)
```
antonioterpin, line 494: `fi = safe_i.astype(corr.dtype)` casts the index to whatever dtype `corr` happens to be (typically float32). openpiv's reference computes `(peaks1_i - 1) * cl` as int*float32 → float32 too, but then assembles the final displacement as `peaks1_i.astype(np.float64) - np.floor(...)` for the non-centroid branches. For centroid both paths stay in float32, so this matches — but the comment block above doesn't note that the JAX port deliberately stays single-precision throughout the centroid arithmetic. Worth a one-line comment so a future reader doesn't try to 'fix' the precision and silently desync from openpiv.

tests/unit/flow/test_openpiv_jax_numerical.py
```diff
@@ -122,6 +128,69 @@ def test_subpixel_displacement_matches_vectorized_reference(
...
 
 
+def test_subpixel_displacement_invalid_method_raises():
+    """An unknown subpixel method raises a clear ValueError."""
+    corr = jnp.ones((1, 8, 8))
+    peaks_i = jnp.array([4])
+    peaks_j = jnp.array([4])
+    with pytest.raises(ValueError, match="Unknown subpixel_method"):
+        subpixel_displacement(corr, peaks_i, peaks_j, subpixel_method="quartic")
+
```
antonioterpin, line 138: The invalid-method test only exercises `subpixel_displacement` directly. Since this PR also adds `subpixel_method` to `extended_search_area_piv` (with new @overloads), there should be a parallel test that `extended_search_area_piv(..., subpixel_method="quartic")` raises — otherwise a regression that drops the validation, or a future refactor that hides the inner call behind another jit/vmap wrapper, would not be caught here.

src/flowgym/flow/open_piv/process.py
```diff
@@ -367,23 +378,45 @@ def subpixel_displacement(
...
         Subpixel displacements (disp_vx, disp_vy).
+
+    Raises:
+        ValueError: If ``subpixel_method`` is not implemented.
     """
+    if subpixel_method not in ("gaussian", "parabolic", "centroid"):
+        raise ValueError(
+            f"Unknown subpixel_method {subpixel_method!r}; expected one of "
+            "'gaussian', 'parabolic', 'centroid'."
+        )
```
antonioterpin, line 419: probably a better approach (also in the other PRs), is to have a type for this params. e.g, and enum or a literal. then, by enforcing typing we could get rid of this.

antonioterpin (changes requested):

src/flowgym/flow/open_piv/process.py
fbanelli: Addressed in 5695a64 via the test route rather than a guard, because the existing numerical parity test pins `~isfinite(disp) == ~isfinite(reference)`: openpiv's parabolic fit divides with the identical unguarded expression, so adding `jnp.where(den != 0, ...)` here would make us return 0.0 where openpiv returns NaN/Inf and *break* parity. Instead I documented the intentional gap inline and added `test_subpixel_parabolic_centroid_unguarded_match_reference`, which feeds explicit peaks (so the result depends only on the division, not argmax) and asserts our NaN/Inf pattern and finite values match openpiv's arithmetic exactly — for a flat stencil (0/0 -> NaN) and an asymmetric one (`cl+cr==2c` -> +/-Inf). A future `jnp.where` "fix" now fails that test. The `invalid` border mask remains what protects production windows.

src/flowgym/flow/open_piv/process.py
fbanelli: Same resolution in 5695a64. The centroid divisor is also unguarded for parity, and I documented the post-normalization failure mode inline: `corr + eps` only keeps the stencil positive while the raw correlation is non-negative, so after `normalized_correlation` an entry below `-eps` can drive `cl + c + cr` toward zero or flip its sign. The parity test's window 2 constructs exactly that (`cl+c+cr ~= 0` via negatives) and confirms we reproduce openpiv's garbage-but-identical absolute position (~1.5e6) rather than diverging.

src/flowgym/flow/open_piv/process.py
fbanelli: Done in 5695a64 — added the one-line note in the centroid branch: the index is cast to `corr.dtype` (float32) so the whole centroid stays single precision, exactly matching openpiv's `int*float32 -> float32` promotion, with an explicit warning not to "fix" it to float64 and silently desync.

tests/unit/flow/test_openpiv_jax_numerical.py
fbanelli: Done in 5695a64 — added `test_pipeline_invalid_subpixel_method_raises`, which calls `extended_search_area_piv(..., subpixel_method="quartic")` and asserts it raises the same `Unknown subpixel_method` `ValueError`. So a refactor that hides the inner `subpixel_displacement` call behind another jit/vmap wrapper and drops the validation is now caught end-to-end, not just at the inner function.

src/flowgym/flow/open_piv/process.py
fbanelli: Agreed. Rather than widen these PRs, I opened #63 to track replacing the runtime `not in {...}` checks with `Literal`/enum types (pinet-style strict typing) so the type checker enforces valid methods statically. For now this PR keeps the explicit `ValueError` with a clear message as a backstop; the typing refactor is tracked there.

antonioterpin (approved):
**Second pass — LGTM.**

All five first-pass points landed correctly (595ad90, 5695a64). Parabolic `(cl-cr)/(2cl-4c+2cr)` and the centroid stencil match openpiv's `vectorized_correlation_to_displacements` line-for-line, and the centroid branch correctly treats its result as an *absolute* position (no peak-index add-back, unlike gaussian/parabolic).

The new `test_subpixel_parabolic_centroid_unguarded_match_reference` genuinely pins the intentionally-unguarded NaN/±Inf behaviour against the openpiv reference (flat stencil → NaN, asymmetric → ±Inf, post-normalization negatives → openpiv's identical ~1.5e6 garbage) — and a future `jnp.where` guard would fail it. Keeping parity over guarding was the right call. Dispatch is sound: eager `ValueError` before any vmap tracing (covered end-to-end), static-string closure traces one branch, default `gaussian` byte-identical to `dev`. Literal/enum typing deferred to #63 is fine.

PR #58, opened 2026-06-06
feat(open_piv): add linear correlation method to fft_correlate_images

fbanelli:
@-

### DIFF
--- a/src/flowgym/flow/open_piv/process.py
+++ b/src/flowgym/flow/open_piv/process.py
@@ -73,6 +73,14 @@ def extended_search_area_piv(
             :func:`fft_correlate_images`, either ``"circular"`` or
             ``"linear"``.
 
+    Note:
+        ``sig2noise_method``, ``subpixel_method`` and ``correlation_method``
+        are Python strings that select code paths, so under ``jax.jit`` they
+        must be marked static (e.g. ``jax.jit(extended_search_area_piv,
+        static_argnames=("window_size", "search_area_size", "overlap",
+        "sig2noise_method", "subpixel_method", "correlation_method"))``)
+        alongside the window-geometry arguments.
+
     Returns:
         Displacement field of shape (batch_size, n_rows, n_cols, 2). If
         ``sig2noise_method`` is set, a tuple of the displacement field and
@@ -224,7 +232,8 @@ def fft_correlate_images(
     """
     if correlation_method not in ("circular", "linear"):
         raise ValueError(
-            f"correlation method {correlation_method} is not implemented"
+            f"Unknown correlation_method {correlation_method!r}; expected "
+            "one of 'circular', 'linear'."
         )
 
     aa = normalize_intensity(aa)
--- a/src/flowgym/flow/open_piv/process.py
+++ b/src/flowgym/flow/open_piv/process.py
@@ -20,6 +20,7 @@ def extended_search_area_piv(
     sig2noise_method: None = None,
     width: int = 2,
     subpixel_method: str = "gaussian",
+    correlation_method: str = "circular",
 ) -> jnp.ndarray: ...
 
 
@@ -33,6 +34,7 @@ def extended_search_area_piv(
     sig2noise_method: str,
     width: int = 2,
     subpixel_method: str = "gaussian",
+    correlation_method: str = "circular",
 ) -> tuple[jnp.ndarray, jnp.ndarray]: ...
 
 
@@ -45,6 +47,7 @@ def extended_search_area_piv(
     sig2noise_method: str | None = None,
     width: int = 2,
     subpixel_method: str = "gaussian",
+    correlation_method: str = "circular",
 ) -> jnp.ndarray | tuple[jnp.ndarray, jnp.ndarray]:
     """Batched PIV cross-correlation algorithm.
 
@@ -66,6 +69,9 @@ def extended_search_area_piv(
         subpixel_method: Sub-pixel peak estimator passed to
             :func:`subpixel_displacement`, one of ``"gaussian"``,
             ``"parabolic"`` or ``"centroid"``.
+        correlation_method: Cross-correlation method passed to
+            :func:`fft_correlate_images`, either ``"circular"`` or
+            ``"linear"``.
 
     Returns:
         Displacement field of shape (batch_size, n_rows, n_cols, 2). If
@@ -134,7 +140,7 @@ def extended_search_area_piv(
 
     # Compute correlation (normalizes the windows internally, matching the
     # reference's normalized_correlation=True path).
-    corr = fft_correlate_images(aa, bb)
+    corr = fft_correlate_images(aa, bb, correlation_method)
 
     # Find peaks and compute displacements. subpixel_method is static, so it
     # is closed over rather than vmapped.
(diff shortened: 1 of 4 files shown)


src/flowgym/flow/open_piv/process.py
```diff
@@ -66,6 +69,9 @@ def extended_search_area_piv(
         subpixel_method: Sub-pixel peak estimator passed to
             :func:`subpixel_displacement`, one of ``"gaussian"``,
             ``"parabolic"`` or ``"centroid"``.
+        correlation_method: Cross-correlation method passed to
+            :func:`fft_correlate_images`, either ``"circular"`` or
+            ``"linear"``.
```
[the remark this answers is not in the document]
fbanelli: Added. The docstring now has a Note that `sig2noise_method`, `subpixel_method` and `correlation_method` are string-selected code paths and must be marked static under `jax.jit` (with an explicit `static_argnames=(...)` example), so jitting the pipeline with `correlation_method="linear"` won't trip users up.

src/flowgym/flow/open_piv/process.py
```diff
@@ -193,25 +199,61 @@ def get_field_shape(
...
     Returns:
         Cross-correlation result (..., height, width).
+
+    Raises:
+        ValueError: If ``correlation_method`` is not implemented.
     """
+    if correlation_method not in ("circular", "linear"):
+        raise ValueError(
+            f"correlation method {correlation_method} is not implemented"
+        )
```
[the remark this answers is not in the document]
fbanelli: Agreed, and taken further: rather than aligning to openpiv's terse wording, both messages are now high-quality and mutually consistent — `fft_correlate_images` raises `Unknown correlation_method 'quadratic'; expected one of 'circular', 'linear'.` and `subpixel_displacement` raises the matching `Unknown subpixel_method ...; expected one of 'gaussian', 'parabolic', 'centroid'.` (the latter improved in #57). They state the offending value and the valid options.

fbanelli:
Addressed both Copilot comments (commit pushed): (1) documented that the string method-selectors must be marked static under `jax.jit`, with an example; (2) improved the `correlation_method` error message to a clear, consistent format (paired with the `subpixel_method` one in #57). Tests/gates green.

src/flowgym/flow/open_piv/process.py
```diff
@@ -193,25 +207,62 @@ def get_field_shape(
...
+            f"Unknown correlation_method {correlation_method!r}; expected "
+            "one of 'circular', 'linear'."
+        )
+
     aa = normalize_intensity(aa)
     bb = normalize_intensity(bb)
 
-    s2 = aa.shape[-2:]
+    s1 = aa.shape[-2:]
+    s2 = bb.shape[-2:]
```
antonioterpin, line 251: This is a silent behaviour change to the pre-existing circular branch: the parent branch had `s2 = aa.shape[-2:]` and divided by that, this PR redefines `s2 = bb.shape[-2:]` and the same `corr / (s2[0]*s2[1])` on line 267 now normalizes by `bb`'s shape. For the in-tree caller `extended_search_area_piv` (`aa` is padded to `bb`'s shape) the two are equal, but any external caller passing `fft_correlate_images` with `aa.shape != bb.shape` on the `circular` path will silently get a different scaling than before. Matching openpiv here is the right call, but please call it out in the PR description / commit message, and consider adding a regression test that pins the denominator (e.g. compares circular against the openpiv reference with `normalized_correlation=True`) so this is locked down rather than incidentally correct.

src/flowgym/flow/open_piv/process.py
```diff
@@ -193,25 +207,62 @@ def get_field_shape(
...
         Cross-correlation result (..., height, width).
+
+    Raises:
+        ValueError: If ``correlation_method`` is not implemented.
     """
+    if correlation_method not in ("circular", "linear"):
+        raise ValueError(
+            f"Unknown correlation_method {correlation_method!r}; expected "
+            "one of 'circular', 'linear'."
+        )
```
antonioterpin, line 245: This `ValueError` only fires when `correlation_method` is a concrete Python string at call time. The `Note:` block on `extended_search_area_piv` instructs callers to mark it `static`, but `fft_correlate_images` is a public symbol re-exported on its own and its docstring does not. If a user does `jax.jit(fft_correlate_images)` and passes a bad method name (or even a valid one), they get an opaque `TypeError: ... interpreting argument ... as an abstract array ... at path correlation_method` from the tracer instead of this clear `ValueError`. I'd add the same static-argname note to this docstring (and ideally to the `Raises:` section, noting it only triggers when the argument is concrete).

src/flowgym/flow/open_piv/process.py
```diff
@@ -193,25 +207,62 @@ def get_field_shape(
...
+    if correlation_method == "linear":
+        # Zero-pad to the reference's fsize = 2**ceil(log2(s1+s2-1)) - 1 and
+        # crop the centred s1-sized region after the inverse transform. The
+        # bit_length form computes the same fsize exactly without float log2.
+        fsize = tuple(
+            (1 << (s1[d] + s2[d] - 2).bit_length()) - 1 for d in (0, 1)
+        )
+        f2a = jnp.conj(jnp.fft.rfft2(aa, s=fsize, axes=(-2, -1)))
+        f2b = jnp.fft.rfft2(bb, s=fsize, axes=(-2, -1))
+        corr = jnp.fft.irfft2(f2a * f2b, axes=(-2, -1)).real
```
antonioterpin, line 269: Worth a brief inline comment that the missing `s=fsize` on `irfft2` is deliberate and reproduces the openpiv quirk: with odd `fsize` (e.g. 63), `rfft2(..., s=63)` produces a last-axis length of 32, and `irfft2` without `s` infers `2*(32-1) = 62`, so `corr` here has shape `(..., fsize[0], fsize[1]-1)` — asymmetric and one short of `fsize` on the last axis — before fftshift and crop. The crop math below still lands the right pixels because the centred window fits inside that shorter axis, but it's non-obvious from a code read and easy to 'fix' into a bug by adding `s=fsize` to match the forward transform. The PR description mentions this; please mirror that comment in the code.

src/flowgym/flow/open_piv/process.py
```diff
@@ -134,7 +148,7 @@ def extended_search_area_piv(
 
     # Compute correlation (normalizes the windows internally, matching the
     # reference's normalized_correlation=True path).
-    corr = fft_correlate_images(aa, bb)
+    corr = fft_correlate_images(aa, bb, correlation_method)
```
antonioterpin: Passing `correlation_method` positionally here (`fft_correlate_images(aa, bb, correlation_method)`) is fine today but couples the two signatures: any future reordering of `fft_correlate_images`'s parameters becomes a silent bug. Since the other string options (`subpixel_method`) are already routed by keyword, please pass this one by keyword as well: `fft_correlate_images(aa, bb, correlation_method=correlation_method)`.

src/flowgym/flow/open_piv/process.py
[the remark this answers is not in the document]
antonioterpin: maybe consider the comment in the other PR about typing

src/flowgym/flow/open_piv/process.py
antonioterpin: Or, typing approach

antonioterpin (changes requested):

src/flowgym/flow/open_piv/process.py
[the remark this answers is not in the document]
fbanelli: Tracked in #63 — same typing theme: `correlation_method`/`subpixel_method`/`sig2noise_method` become `Literal`/enum so the static-arg note + runtime check can largely be replaced by the type checker. Kept the runtime `ValueError` + docstring note here for now.

src/flowgym/flow/open_piv/process.py
fbanelli: Same — folded into the typing follow-up in #63.

src/flowgym/flow/open_piv/process.py
fbanelli: Done in 62c1af9. Called it out in the commit message and appended a "Behavior note" to the PR description, and added `test_fft_correlate_circular_matches_reference`: it compares the raw circular output against `pyprocess.fft_correlate_images(..., normalized_correlation=True)` across several window sizes, pinning the `s2 = bb.shape` denominator. As you note, matching openpiv is the intended behaviour; the in-tree path is unchanged (aa is padded to bb), and the denominator can no longer silently drift back to `aa.shape`.

src/flowgym/flow/open_piv/process.py
fbanelli: Done in 62c1af9. Added the static-argname Note to `fft_correlate_images`' own docstring (e.g. `static_argnames="correlation_method"`), and amended the `Raises:` line to state the `ValueError` only fires when the argument is a concrete Python string — under `jax.jit` with a traced argument you instead get the opaque tracer `TypeError`. (The broader "make it a Literal so this is enforced statically" idea is tracked in #63.)

src/flowgym/flow/open_piv/process.py
fbanelli: Done in 62c1af9 — mirrored the PR-description explanation as an inline comment above the `irfft2` call: with odd `fsize` (e.g. 63), `rfft2` gives a last-axis length of 32 and `irfft2` without `s` infers `2*(32-1) = 62`, so `corr`'s last axis is `fsize-1`; the centred crop still lands the right pixels, and the comment explicitly warns against adding `s=fsize` to "match" the forward transform.

src/flowgym/flow/open_piv/process.py
fbanelli: Done in 62c1af9 — now `fft_correlate_images(aa, bb, correlation_method=correlation_method)`, so a future reordering of `fft_correlate_images`'s parameters can't silently miswire it.

tests/unit/flow/test_openpiv_jax_numerical.py
```diff
@@ -271,6 +271,133 @@ def test_pipeline_subpixel_method_matches_reference(
...
+        (overlap, overlap),
+    )[0]
+    bb = sliding_window_array(
+        jnp.asarray(frame_b)[None],
+        (window_size, window_size),
+        (overlap, overlap),
+    )[0]
+
+    got = np.asarray(
+        fft_correlate_images(aa, bb, correlation_method="circular")
```
antonioterpin, line 329: This regression test doesn't actually pin the `s2 = bb.shape` choice: `aa` and `bb` are both built with `(window_size, window_size)`, so `aa.shape == bb.shape` and the denominator `s2[0]*s2[1]` is identical whether you index `bb.shape` or `aa.shape` — a silent revert to the old `s2 = aa.shape[-2:]` would produce byte-identical output and still pass. The test pins openpiv-parity for the equal-shape case but not the bb-vs-aa fix it documents. Add a case with `aa.shape != bb.shape` (e.g. call `fft_correlate_images` directly on a deliberately mismatched pair, or assert the normalized output equals `raw/(bb.H*bb.W)` with `bb` smaller than `aa`).

antonioterpin (changes requested):
**Second pass — I rebased this onto current #57 (now in `dev`) and retargeted the base to `dev`; one test fix still needed before merge.**

The stale-base problem is resolved: I rebased `feat/openpiv-linear-correlation` onto the updated `feat/openpiv-subpixel-methods` tip (force-pushed `62c1af9` → `6562de9`) and retargeted this PR to `dev`. The diff is now clean and purely additive (+223/−11, exactly #58's own linear-correlation work) — #57's improved `subpixel_method` error message and its two parity tests (`test_pipeline_invalid_subpixel_method_raises`, `test_subpixel_parabolic_centroid_unguarded_match_reference`) are preserved, and the branch is MERGEABLE.

The technical work is right: the linear path matches openpiv's `correlation_method="linear"` to ~2e-7 (the `fsize` bit_length identity, the centred crop, and the deliberate no-`s` `irfft2` quirk all check out), and `s2 = bb.shape[-2:]` correctly fixes the latent `s2 = aa.shape` external-caller bug (matches openpiv L747).

**Remaining requested change (inline):** the circular regression test doesn't actually lock the bb-vs-aa denominator it claims to.

tests/unit/flow/test_openpiv_jax_numerical.py
antonioterpin: Done in 9936769. Added `test_fft_correlate_linear_pins_bb_denominator`: it feeds a deliberately mismatched pair (`aa` 24×24, `bb` 16×16) through the **linear** path — the only branch that admits unequal shapes, since `circular`'s `rfft2` multiply broadcast-fails on mismatched windows — so dividing by `aa`'s area (576) instead of `bb`'s (256) rescales every pixel by 256/576 and breaks parity with the openpiv reference.

`s2 = bb.shape[-2:]` and the `corr / (s2[0]*s2[1])` division are defined once and shared by both branches, so locking it on the linear path also locks `circular`. Verified by reverting `s2 = bb.shape → aa.shape`: the new test fails (all 1728 pixels differ, max abs diff 0.14, and a 200-seed sweep keeps the min max-diff at 0.129 — well above `atol=1e-5`) while the equal-shape circular parity tests still pass.

Also corrected the circular test's docstring, which had overclaimed that it pinned the denominator; it now scopes itself to equal-shape parity and cross-references the new test.

PR #59, opened 2026-06-06
feat(open_piv): support rectangular (non-square) windows

fbanelli:
Stacked on #58 → #57 → `dev` (shares `extended_search_area_piv`; bases collapse as the stack merges).

## What

`extended_search_area_piv` squared every size argument and asserted int-only inputs — the long-standing `# TODO: extend to handle non-square windows`. The downstream helpers (`sliding_window_array`, `get_field_shape`, `fft_correlate_images`, `subpixel_displacement`) already work per-axis, so the only barrier was the top-level API.

- Accept `int` **or** `(height, width)` tuples for `window_size`, `search_area_size`, `overlap` via a small `_as_pair` helper; normalize to tuples; validate `search ≥ window` and `overlap < search` per axis.
- The extended-search masking keeps openpiv's **lexicographic tuple comparison** (`search_area_size > window_size`) verbatim — under the per-axis `search ≥ window` constraint this means "extended whenever the search area is larger on at least one axis", matching the reference.
- `int` inputs still produce the identical square result (no behaviour change for existing callers; the `OpenPIVJAXEstimator` keeps passing ints).

## Tests (1-to-1 openpiv parity, established layout)

- `test_openpiv_jax_numerical.py`: `test_pipeline_rectangular_windows_matches_reference` vs `pyprocess.extended_search_area_piv` across 4 rectangular configs (standard, transposed, extended on one axis, extended on both).
- `test_openpiv_jax_jit.py`: pipeline traces with static rectangular tuple geometry.
- `test_openpiv_jax.py`: `int` window size equals the `(n, n)` tuple form.

Rectangular configs match the reference to ~1e-6; NaN/invalid positions agree exactly.

## Verification

121 openpiv tests pass locally (CPU; up from 115). ruff (pinned 0.14.1), pydoclint, basedpyright clean on the changed files.

### DIFF
--- a/src/flowgym/flow/open_piv/process.py
+++ b/src/flowgym/flow/open_piv/process.py
@@ -10,6 +10,32 @@ from flowgym.flow.process import img_resize
 from flowgym.utils import DEBUG
 
 
+def _as_int(value: int) -> int:
+    """Coerce an integer-like scalar to ``int``, rejecting non-integers.
+
+    Window/overlap normalization runs host-side (not in the traced path),
+    so this guard is always on rather than ``DEBUG``-gated: a non-integer
+    such as ``32.5`` would otherwise silently truncate to ``32`` and yield
+    a wrong-geometry PIV field instead of a clear error. Integer-valued
+    floats (``32.0``) and ``numpy`` integers pass through unchanged.
+
+    Args:
+        value: An integer-like scalar.
+
+    Returns:
+        The value as a Python ``int``.
+
+    Raises:
+        ValueError: If the value is not integer-like.
+    """
+    ivalue = int(value)
+    if ivalue != value:
+        raise ValueError(
+            f"Expected an integer-like window/overlap size, got {value!r}."
+        )
+    return ivalue
+
+
 def _as_pair(value: int | tuple[int, int]) -> tuple[int, int]:
     """Normalize an int or (height, width) pair to a 2-tuple of ints.
 
@@ -17,7 +43,8 @@ def _as_pair(value: int | tuple[int, int]) -> tuple[int, int]:
     given as a single int while rectangular windows use an explicit pair.
     A tuple/list is taken as the pair; any other value is treated as a
     scalar and duplicated, so integer-like scalars such as ``numpy.int32``
-    work too.
+    work too. Each element is coerced via :func:`_as_int`, so non-integer
+    sizes raise rather than silently truncating.
 
     Args:
         value: Either an integer-like scalar (square) or a ``(height,
@@ -32,8 +59,8 @@ def _as_pair(value: int | tuple[int, int]) -> tuple[int, int]:
                 "Expected an int or a (height, width) pair, got "
                 f"length-{len(value)} {value!r}."
             )
-        return (int(value[0]), int(value[1]))
-    return (int(value), int(value))
+        return (_as_int(value[0]), _as_int(value[1]))
+    return (_as_int(value), _as_int(value))
 
 
 @overload
@@ -175,9 +202,9 @@ def extended_search_area_piv(
     # this is equivalent to "search is larger on at least one axis": given
     # s[0] >= w[0] and s[1] >= w[1], if s != w then either s[0] > w[0] (lex
     # true via the first element) or s[0] == w[0] and s[1] > w[1] (lex true
-    # via the second). Without that precondition the equivalence breaks --
-    # e.g. search=(16, 64) vs window=(32, 16) would lex-trigger this branch
-    # while being invalid input.
+    # via the second). Without it the equivalence breaks the other way:
+    # search=(16, 64) vs window=(32, 16) is larger on axis 1 but lex-False
+    # (axis 0 decides), so this branch is wrongly skipped.
     if search_area_size_tuple > window_size_tuple:
         aa = normalize_intensity(aa)
         bb = normalize_intensity(bb)
--- a/src/flowgym/flow/open_piv/process.py
+++ b/src/flowgym/flow/open_piv/process.py
@@ -27,6 +27,11 @@ def _as_pair(value: int | tuple[int, int]) -> tuple[int, int]:
         The value as a ``(height, width)`` tuple of ints.
     """
     if isinstance(value, (tuple, list)):
+        if DEBUG:
+            assert len(value) == 2, (
+                "Expected an int or a (height, width) pair, got "
+                f"length-{len(value)} {value!r}."
+            )
         return (int(value[0]), int(value[1]))
     return (int(value), int(value))
 
@@ -101,6 +106,14 @@ def extended_search_area_piv(
         static_argnames=("window_size", "search_area_size", "overlap",
         "sig2noise_method", "subpixel_method", "correlation_method"))``)
         alongside the window-geometry arguments.
+
+        Input validation is opt-in: the geometry constraints
+        (``search_area_size >= window_size`` and
+        ``overlap < search_area_size`` per axis) and the ``(height, width)``
+        shape of tuple arguments are checked only under the module ``DEBUG``
+        flag. With ``DEBUG`` disabled (the default) malformed geometry is not
+        rejected here -- this is intentional for runtime/jit leanness, not a
+        missing check.
 
     Returns:
         Displacement field of shape (batch_size, n_rows, n_cols, 2). If
--- a/src/flowgym/flow/open_piv/process.py
+++ b/src/flowgym/flow/open_piv/process.py
@@ -10,13 +10,30 @@ from flowgym.flow.process import img_resize
 from flowgym.utils import DEBUG
 
 
+def _as_pair(value: int | tuple[int, int]) -> tuple[int, int]:
+    """Normalize an int or (height, width) pair to a 2-tuple of ints.
+
+    Mirrors openpiv's scalar-to-tuple reshaping so square windows can be
+    given as a single int while rectangular windows use an explicit pair.
+
+    Args:
+        value: Either a single int (square) or an (height, width) pair.
+
+    Returns:
+        The value as a ``(height, width)`` tuple of ints.
+    """
+    if isinstance(value, int):
+        return (value, value)
+    return (int(value[0]), int(value[1]))
+
+
 @overload
 def extended_search_area_piv(
     img1: jnp.ndarray,
     img2: jnp.ndarray,
-    window_size: int,
-    search_area_size: int,
-    overlap: int,
+    window_size: int | tuple[int, int],
+    search_area_size: int | tuple[int, int],
+    overlap: int | tuple[int, int],
     sig2noise_method: None = None,
     width: int = 2,
     subpixel_method: str = "gaussian",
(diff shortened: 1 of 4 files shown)


src/flowgym/flow/open_piv/process.py
```diff
@@ -10,13 +10,30 @@
...
+
+    Args:
+        value: Either a single int (square) or an (height, width) pair.
+
+    Returns:
+        The value as a ``(height, width)`` tuple of ints.
+    """
+    if isinstance(value, int):
+        return (value, value)
+    return (int(value[0]), int(value[1]))
```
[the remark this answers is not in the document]
fbanelli: Real regression — fixed. `_as_pair` now treats tuples/lists as the pair and any other value as a scalar via `int(value)`, so `numpy.int32`/`numpy.int64` work again (they previously fell into the subscript branch and raised `'numpy.int32' object is not subscriptable'`). Added a numpy-integer regression test.

fbanelli:
Addressed the Copilot review: `_as_pair` now accepts integer-like scalars (e.g. `numpy.int32`) by treating only tuples/lists as the pair, fixing the regression vs the old `(value, value)` duplication. Added a numpy-integer regression test. Gates green.

src/flowgym/flow/open_piv/process.py
```diff
@@ -86,18 +113,14 @@ def extended_search_area_piv(
...
-    overlap_tuple = (overlap, overlap)
-    search_area_size_tuple = (search_area_size, search_area_size)
+        assert (
+            search_area_size_tuple[0] >= window_size_tuple[0]
+            and search_area_size_tuple[1] >= window_size_tuple[1]
+        ), "Search area size must be >= window size on both axes"
+        assert (
+            overlap_tuple[0] < search_area_size_tuple[0]
+            and overlap_tuple[1] < search_area_size_tuple[1]
+        ), "Overlap must be smaller than the search area size on both axes"
```
antonioterpin, line 170: Document that input validation is opt-in (in general, I think we never wrote it in docs and agents pick it up as a bug)

src/flowgym/flow/open_piv/process.py
```diff
@@ -126,7 +149,10 @@ def extended_search_area_piv(
     # and normalizes BEFORE masking so the zeroed border does not pollute the
     # per-window mean/std. fft_correlate_images then normalizes once more, so
     # the extended-search branch is normalized twice exactly as in openpiv.
-    if search_area_size > window_size:
+    # Lexicographic tuple comparison, matching openpiv exactly. Under the
+    # per-axis search >= window constraint this activates whenever the search
+    # area is larger than the window on at least one axis.
+    if search_area_size_tuple > window_size_tuple:
```
antonioterpin, line 207: The comment says lex `>` 'activates whenever the search area is larger than the window on at least one axis,' which is true — but only because of the per-axis `search >= window` constraint asserted above. Worth spelling out the proof in one line, because the equivalence is not obvious and the constraint that makes it hold is in a `DEBUG`-gated assert (see other comment): given `s[0] >= w[0]` and `s[1] >= w[1]`, if `s != w` then either `s[0] > w[0]` (lex true via first element) or `s[0] == w[0] and s[1] > w[1]` (lex true via second). Without the per-axis precondition, e.g. `search=(16, 64)` vs `window=(32, 16)` would mis-trigger the branch in the lex sense while being invalid input.

tests/unit/flow/test_openpiv_jax_numerical.py
```diff
@@ -281,6 +281,66 @@ def test_pipeline_linear_correlation_matches_reference(
...
+    np.testing.assert_array_equal(
+        ~np.isfinite(u_jax), ~np.isfinite(np.asarray(u_ref))
+    )
+    finite = np.isfinite(u_jax) & np.isfinite(np.asarray(u_ref))
+    assert finite.any()
+    np.testing.assert_allclose(
+        u_jax[finite], np.asarray(u_ref)[finite], atol=1e-2
+    )
+    np.testing.assert_allclose(
+        v_jax[finite], np.asarray(v_ref)[finite], atol=1e-2
```
antonioterpin, line 492: The new tests cover the numerical happy path well but don't exercise any of the validation that this PR explicitly adds (`search >= window` per axis, `overlap < search` per axis). Please add at least two failure-mode tests:

- `search_area_size=(16, 16)` with `window_size=(32, 32)` — must raise (currently passes silently in production because the assert is under `DEBUG`).
- `overlap=(16, 16)` with `search_area_size=(16, 16)` — must raise.

Also worth one test for a mismatched-length tuple (e.g. `window_size=(32, 16, 8)`) to lock down `_as_pair`'s contract. As written, these inputs silently produce wrong results or low-level downstream errors rather than the clear `ValueError` the openpiv reference raises.

antonioterpin (changes requested):

src/flowgym/flow/open_piv/process.py
fbanelli: Done in f365df9 — expanded the comment with the one-line proof: under the per-axis `search >= window` precondition, `s != w` implies either `s[0] > w[0]` (lex true via the first element) or `s[0] == w[0] and s[1] > w[1]` (lex true via the second), so the lexicographic `>` is exactly "larger on at least one axis". Also noted the counterexample that breaks the equivalence without the precondition (`search=(16, 64)` vs `window=(32, 16)`), and that the precondition itself is the DEBUG-gated assert above — which ties into the validation-opt-in discussion on the sibling thread.

src/flowgym/flow/open_piv/process.py
fbanelli: Done in 45edcc4 — added a `Note:` to `extended_search_area_piv` stating that input validation (the `search >= window` / `overlap < search` geometry constraints and the `(height, width)` tuple shape) is opt-in, enforced only under the module `DEBUG` flag, and that with `DEBUG` off malformed geometry passing silently is intentional rather than a missing check. I also extended `_as_pair` to flag malformed (len != 2) tuples under `DEBUG` (it previously truncated 3-tuples silently), keeping the same opt-in convention.

tests/unit/flow/test_openpiv_jax_numerical.py
fbanelli: Addressed in 45edcc4, but kept the validation **opt-in** rather than promoting it to always-on `ValueError`s, to honour the maintainer note on the sibling thread (validation is opt-in by design; always-on checks shouldn't be added). Concretely: added tests that enable `DEBUG` and assert the checks fire — `search_area_size=(16,16)` with `window_size=(32,32)`, `overlap == search_area_size`, and a mismatched 3-tuple through `_as_pair` — plus a docstring note documenting the opt-in behaviour so it isn't re-flagged as a bug. If you'd rather these become always-on `ValueError`s (openpiv-style) at the cost of the opt-in convention, happy to flip them — just confirm and I'll promote the asserts and adjust the tests.

src/flowgym/flow/open_piv/process.py
```diff
@@ -126,7 +163,15 @@ def extended_search_area_piv(
...
     # per-window mean/std. fft_correlate_images then normalizes once more, so
     # the extended-search branch is normalized twice exactly as in openpiv.
-    if search_area_size > window_size:
+    # Lexicographic tuple comparison, matching openpiv exactly. Under the
+    # per-axis `search >= window` precondition (asserted above under DEBUG)
+    # this is equivalent to "search is larger on at least one axis": given
+    # s[0] >= w[0] and s[1] >= w[1], if s != w then either s[0] > w[0] (lex
+    # true via the first element) or s[0] == w[0] and s[1] > w[1] (lex true
+    # via the second). Without that precondition the equivalence breaks --
+    # e.g. search=(16, 64) vs window=(32, 16) would lex-trigger this branch
```
antonioterpin: The counterexample direction is backwards: `(16, 64) > (32, 16)` is actually **False** (lex compares 16 < 32 first), so it does NOT trigger this branch. The example does correctly show the equivalence breaking, but in the opposite sense — search is larger on one axis (64 > 16) yet lex returns False, so the extended branch is wrongly *skipped*. Reword to e.g. "would be skipped by this lex comparison despite search being larger on axis 1".

src/flowgym/flow/open_piv/process.py
```diff
@@ -10,13 +10,39 @@
...
+        The value as a ``(height, width)`` tuple of ints.
+    """
+    if isinstance(value, (tuple, list)):
+        if DEBUG:
+            assert len(value) == 2, (
+                "Expected an int or a (height, width) pair, got "
+                f"length-{len(value)} {value!r}."
+            )
+        return (int(value[0]), int(value[1]))
+    return (int(value), int(value))
```
antonioterpin: `int(value)` silently truncates a non-integer float (`32.5` → `32`, likewise `(32.5, 16)` on the tuple branch) with no guard even under DEBUG — only tuple *length* is asserted. Since this normalization is host-side Python (not in the traced path), I'd make this an **always-on** `ValueError` for non-integer-like inputs (or at minimum a DEBUG assert that `int(value) == value`). Silent truncation yields a wrong-geometry PIV field rather than an error.

antonioterpin (changes requested):

src/flowgym/flow/open_piv/process.py
antonioterpin: Done in 4a006de. Added `_as_int`, an **always-on** `ValueError` guard (host-side Python, not in the traced path); `_as_pair` now routes every element through it, so `32.5` / `(32.5, 16)` raise `Expected an integer-like window/overlap size, got 32.5.` instead of silently truncating to `32` and yielding a wrong-geometry field. Integer-valued floats (`32.0`) and `numpy` integers still pass — `test_numpy_integer_window_sizes` stays green. Tests cover scalar + both tuple positions rejecting, plus an integer-float acceptance case.

src/flowgym/flow/open_piv/process.py
antonioterpin: Done in 4a006de. Reworded — `(16, 64) > (32, 16)` is `False` (axis 0 decides, `16 < 32`), so the branch is wrongly *skipped* despite the search being larger on axis 1, not mis-triggered. New comment: "search=(16, 64) vs window=(32, 16) is larger on axis 1 but lex-False (axis 0 decides), so this branch is wrongly skipped."

src/flowgym/flow/open_piv/process.py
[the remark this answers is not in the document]
antonioterpin: Reinforced in 4a006de: `_as_pair` now coerces each element via `_as_int`, which accepts any integer-like scalar (incl. `numpy.int32`/`int64`) and rejects non-integers. Resolving.

src/flowgym/flow/open_piv/process.py
```diff
@@ -81,11 +134,25 @@ def extended_search_area_piv(
         "sig2noise_method", "subpixel_method", "correlation_method"))``)
         alongside the window-geometry arguments.
 
+        Input validation is opt-in: the geometry constraints
+        (``search_area_size >= window_size`` and
+        ``overlap < search_area_size`` per axis) and the ``(height, width)``
+        shape of tuple arguments are checked only under the module ``DEBUG``
+        flag. With ``DEBUG`` disabled (the default) malformed geometry is not
+        rejected here -- this is intentional for runtime/jit leanness, not a
+        missing check.
```
antonioterpin: remove -- and everything after

antonioterpin: Done in dca27a3 — trimmed to "...malformed geometry is not rejected here." (dropped the `-- this is intentional...` justification).

PR #60, opened 2026-06-06
feat(open_piv): window deformation + fixed-grid multipass

fbanelli:
Branches off `dev` independently of the open #57/#58/#59 stack (new standalone functions; no overlap with their signature changes).

## What

Ports openpiv window deformation to JAX and builds the iterative multipass on top — the **main accuracy lever** for non-uniform flows, which the JAX pipeline previously lacked entirely.

1. **`deform_windows`** — resamples an image onto the grid deformed by a coarse displacement field (the core deformation primitive).
2. **`multipass_deform`** — fixed-resolution iterative refinement: estimate → deform second image by the current estimate → re-correlate for the residual → accumulate.

I went in expecting cubic-spline warping to block this; both the deformation *and* the multipass loop reproduce openpiv exactly at linear interpolation.

## Exact-parity details

- **`deform_windows`**: field upsampling is bilinear-with-clamping (openpiv's `RectBivariateSpline` degree 1 extrapolates as a constant outside the window-centre grid — matched to ~4e-16); image resampling uses `map_coordinates` order 1, `mode='nearest'`, matching `scipy.ndimage`.
- **`multipass_deform`**: follows openpiv's `deformation_method='second image'` recipe (deform with `-v` so the second image is sampled at `(y + v, x + u)`; accumulate the residual). Pinned against a reference loop assembled from openpiv's own primitives (`windef.deform_windows` + `pyprocess.extended_search_area_piv`) — matches to ~1e-3.

## Scope / honest limitations (documented)

- Linear interpolation only; openpiv's *default* cubic field interpolation (`RectBivariateSpline` degree 3) has no exact JAX equivalent.
- `multipass_deform` holds the grid fixed across passes. Grid refinement (changing window sizes per pass) and statistical outlier replacement between passes are out of scope — the caller can apply `replace_outliers`.

## Tests (new gated file `test_openpiv_jax_deform.py`)

- `deform_windows`: parity vs `windef.deform_windows(order=order2=1)` across grid/flow configs, zero-flow identity, uniform-shift, jit.
- `multipass_deform`: parity vs the openpiv-primitive reference loop (n_passes 2/3/4), n_passes=1 == single pass, accuracy-does-not-worsen on a known non-uniform flow, jit.

13 tests pass locally (CPU). ruff (pinned 0.14.1), pydoclint, basedpyright clean.

### DIFF
--- a/src/flowgym/flow/open_piv/process.py
+++ b/src/flowgym/flow/open_piv/process.py
@@ -900,3 +900,85 @@ def deform_windows(
     return jax.scipy.ndimage.map_coordinates(
         frame, [pixel_y - vt, pixel_x + ut], order=1, mode="nearest"
     )
+
+
+def multipass_deform(
+    img1: jnp.ndarray,
+    img2: jnp.ndarray,
+    window_size: int,
+    search_area_size: int,
+    overlap: int,
+    n_passes: int,
+) -> jnp.ndarray:
+    """Iterative window-deformation PIV at a fixed grid resolution.
+
+    Runs the standard window-deformation refinement: estimate the
+    displacement, deform the second image toward the first by the current
+    estimate, re-correlate to obtain the residual, and accumulate. This is
+    the main accuracy lever for non-uniform flows that a single pass
+    under-resolves. The grid resolution (window/search/overlap) is held
+    fixed across passes, matching openpiv's ``deformation_method="second
+    image"`` recipe (the second image is deformed with ``-v`` so the sample
+    grid is ``(y + v, x + u)``).
+
+    NaN windows are replaced with zero before deformation and accumulation;
+    statistical outlier replacement between passes (openpiv's
+    ``replace_outliers``) is left to the caller.
+
+    Args:
+        img1: First image batch of shape (B, H, W).
+        img2: Second image batch of shape (B, H, W).
+        window_size: Interrogation window size.
+        search_area_size: Search area size.
+        overlap: Overlap between windows.
+        n_passes: Number of passes (``1`` reduces to a single correlation).
+
+    Returns:
+        Displacement field of shape (B, n_rows, n_cols, 2).
+    """
+    if DEBUG:
+        assert img1.ndim == 3, (
+            f"Image must be (batch, height, width), instead {img1.shape}"
+        )
+        assert isinstance(n_passes, int) and n_passes >= 1, (
+            "n_passes must be a positive integer."
+        )
+
+    flow = extended_search_area_piv(
+        img1,
+        img2,
+        window_size=window_size,
+        search_area_size=search_area_size,
+        overlap=overlap,
+    )
+    if n_passes <= 1:
+        return flow
+
+    height, width = img1.shape[1], img1.shape[2]
+    search_tuple = (search_area_size, search_area_size)
+    overlap_tuple = (overlap, overlap)
+    n_rows, n_cols = get_field_shape(
+        (height, width), search_tuple, overlap_tuple
+    )
+    xs, ys = get_rect_coordinates((height, width), search_tuple, overlap_tuple)
+    x_grid = xs.reshape(n_rows, n_cols)
+    y_grid = ys.reshape(n_rows, n_cols)
+
+    def deform_batch(frame, u, v):
+        # "second image" method: deform with -v so frame_b is sampled at
+        # (y + v, x + u), the current displacement estimate.
+        return deform_windows(frame, x_grid, y_grid, u, -v)
+
+    flow = jnp.nan_to_num(flow)
+    for _ in range(n_passes - 1):
+        img2_def = jax.vmap(deform_batch)(img2, flow[..., 0], flow[..., 1])
+        residual = extended_search_area_piv(
+            img1,
+            img2_def,
+            window_size=window_size,
+            search_area_size=search_area_size,
+            overlap=overlap,
+        )
+        flow = flow + jnp.nan_to_num(residual)
+
+    return flow
--- a/src/flowgym/flow/open_piv/process.py
+++ b/src/flowgym/flow/open_piv/process.py
@@ -1,5 +1,6 @@
 """Module for OpenPIV processing in JAX."""
 
+from collections.abc import Callable
 from typing import overload
 
 import jax
@@ -819,7 +820,7 @@ def upsample_flow(
     return jnp.stack((flows_x_resized, flows_y_resized), axis=-1)
 
 
-def deform_windows(
+def _deform_windows(
     frame: jnp.ndarray,
     x: jnp.ndarray,
     y: jnp.ndarray,
@@ -827,6 +828,11 @@ def deform_windows(
     v: jnp.ndarray,
 ) -> jnp.ndarray:
     """Deform an image by the displacement field of a previous PIV pass.
+
+    Internal helper for :func:`multipass_deform`; kept private because ``v``
+    follows openpiv's image-y-down convention (the image is sampled at
+    ``(y - v, x + u)``), which is a footgun for direct callers. Use
+    :func:`multipass_deform`, which owns the sign handling.
 
     JAX port of openpiv's ``windef.deform_windows`` at linear interpolation
     (``interpolation_order = interpolation_order2 = 1``). The coarse
@@ -857,7 +863,19 @@ def deform_windows(
 
     Returns:
         The deformed image of shape (height, width).
+
+    Raises:
+        ValueError: If the window-centre grid has fewer than 2 points on an
+            axis (no defined spacing).
     """
+    # Always-on (DEBUG-independent): with DEBUG off, JAX clamps the out-of-range
+    # y1[1]/x1[1] index so a 1xN/Nx1 grid would silently return a non-finite
+    # image instead of raising. The grid shape is static, so fail fast here.
+    if x.shape[0] < 2 or x.shape[1] < 2:
+        raise ValueError(
+            "Window-centre grid needs >= 2 points on each axis to define a "
+            f"spacing; got grid shape {x.shape}."
+        )
     if DEBUG:
         assert frame.ndim == 2, (
             f"Frame must be 2D (height, width), instead {frame.shape}"
@@ -867,10 +885,6 @@ def deform_windows(
         )
         assert u.shape == x.shape and v.shape == x.shape, (
             "u and v must match the window-centre grid shape."
-        )
-        assert x.shape[0] >= 2 and x.shape[1] >= 2, (
-            "Window-centre grid needs >= 2 points on each axis to define a "
-            f"spacing; got grid shape {x.shape}."
         )
 
     frame = frame.astype(jnp.float32)
@@ -913,6 +927,7 @@ def multipass_deform(
     search_area_size: int,
     overlap: int,
     n_passes: int,
+    between_passes: Callable[[jnp.ndarray, int], jnp.ndarray] | None = None,
 ) -> jnp.ndarray:
     """Iterative window-deformation PIV at a fixed grid resolution.
 
@@ -930,8 +945,9 @@ def multipass_deform(
     :func:`extended_search_area_piv`), while ``n_passes >= 2`` zeroes them with
     ``nan_to_num`` before deformation and accumulation, matching openpiv's
     iterative recipe. Callers that mask on ``isfinite`` must account for this
-    switch. Without between-pass outlier replacement the estimate can drift
-    (see #65); ``replace_outliers`` between passes is left to the caller.
+    switch. Without between-pass outlier replacement the estimate drifts for
+    ``n_passes > 2`` (it peaks at pass 2; see #65); pass ``between_passes`` to
+    clean the field between passes (e.g. an outlier replacement) to avoid this.
 
     Args:
         img1: First image batch of shape (B, H, W).
@@ -940,13 +956,18 @@ def multipass_deform(
         search_area_size: Search area size.
         overlap: Overlap between windows.
         n_passes: Number of passes (``1`` reduces to a single correlation).
+        between_passes: Optional ``(flow, pass_index) -> flow`` callback applied
+            to the accumulated field after each deformation pass (``pass_index``
+            counts from 1), before the next pass deforms by it. Use it to inject
+            outlier replacement and suppress the ``n_passes > 2`` drift. The
+            default ``None`` leaves the field untouched (openpiv parity).
 
     Note:
         ``window_size``, ``search_area_size``, ``overlap`` and ``n_passes``
         are baked in at trace time (the ``range(n_passes - 1)`` loop and
         ``get_field_shape``), so under ``jax.jit`` they must be marked static
         (e.g. ``static_argnames=("window_size", "search_area_size",
-        "overlap", "n_passes")``).
+        "overlap", "n_passes")``); ``between_passes`` must be static too.
 
     Returns:
         Displacement field of shape (B, n_rows, n_cols, 2).
@@ -982,10 +1003,10 @@ def multipass_deform(
     def deform_batch(frame, u, v):
         # "second image" method: deform with -v so frame_b is sampled at
         # (y + v, x + u), the current displacement estimate.
-        return deform_windows(frame, x_grid, y_grid, u, -v)
+        return _deform_windows(frame, x_grid, y_grid, u, -v)
 
     flow = jnp.nan_to_num(flow)
-    for _ in range(n_passes - 1):
+    for i in range(n_passes - 1):
         img2_def = jax.vmap(deform_batch)(img2, flow[..., 0], flow[..., 1])
         residual = extended_search_area_piv(
             img1,
@@ -995,5 +1016,7 @@ def multipass_deform(
             overlap=overlap,
         )
         flow = flow + jnp.nan_to_num(residual)
+        if between_passes is not None:
+            flow = between_passes(flow, i + 1)
 
     return flow
--- a/src/flowgym/flow/open_piv/process.py
+++ b/src/flowgym/flow/open_piv/process.py
@@ -845,9 +845,8 @@ def deform_windows(
       ``map_coordinates`` linear interpolation and ``"nearest"`` border mode
       as the reference.
 
-    Only the linear case is reproduced: openpiv's default cubic field
-    interpolation (``RectBivariateSpline`` degree 3) has no exact JAX
-    equivalent.
+    Only linear field interpolation is implemented; cubic
+    (``RectBivariateSpline`` degree 3) is tracked in #64.
 
     Args:
         frame: Single image of shape (height, width).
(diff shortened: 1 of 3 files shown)


src/flowgym/flow/open_piv/process.py
```diff
@@ -584,3 +584,168 @@ def upsample_flow(
...
+
+    frame = frame.astype(jnp.float32)
+    height, width = frame.shape
+
+    # Window-centre grid is uniformly spaced (overlap-defined), so a pixel
+    # coordinate maps to a fractional field index by an affine transform.
+    y1 = y[:, 0]
+    x1 = x[0, :]
+    dy = y1[1] - y1[0]
+    dx = x1[1] - x1[0]
```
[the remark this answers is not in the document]
fbanelli: Both addressed: added a DEBUG assert that the window-centre grid has >= 2 points on each axis (a 1xN/Nx1 field has no defined spacing), and cast the grid vectors to float32 so the index math no longer promotes to float64.

fbanelli:
Addressed the Copilot review: `deform_windows` now asserts the window-centre grid has >= 2 points per axis (fail-fast on 1xN/Nx1 fields) and casts the grid vectors to float32 to avoid float64 promotion. Gates green.

src/flowgym/flow/open_piv/process.py
```diff
@@ -584,3 +584,173 @@ def upsample_flow(
...
+
+    flow = extended_search_area_piv(
+        img1,
+        img2,
+        window_size=window_size,
+        search_area_size=search_area_size,
+        overlap=overlap,
+    )
+    if n_passes <= 1:
+        return flow
```
antonioterpin, line 991: Inconsistent NaN handling between fast-path and iterative path. When `n_passes <= 1` the first-pass `flow` is returned with NaNs intact (failed correlations propagate to the caller, matching `extended_search_area_piv`). When `n_passes >= 2`, line 744 does `flow = jnp.nan_to_num(flow)` and the returned value never re-introduces NaN. So switching from `n_passes=1` to `n_passes=2` silently changes the failure semantics: callers that mask on `isfinite` will suddenly see zeros instead of NaNs in failed windows, contaminating downstream statistics (mean/median) without any indication. Either (a) also `nan_to_num` in the fast path for consistency, (b) preserve the NaN mask across passes and re-apply at the end, or (c) document this difference loudly in the docstring. The test `test_multipass_single_pass_equals_extended_search` actually pins the current asymmetric behavior.

src/flowgym/flow/open_piv/process.py
```diff
@@ -584,3 +584,173 @@ def upsample_flow(
...
+        (height, width), search_tuple, overlap_tuple
+    )
+    xs, ys = get_rect_coordinates((height, width), search_tuple, overlap_tuple)
+    x_grid = xs.reshape(n_rows, n_cols)
+    y_grid = ys.reshape(n_rows, n_cols)
+
+    def deform_batch(frame, u, v):
+        # "second image" method: deform with -v so frame_b is sampled at
+        # (y + v, x + u), the current displacement estimate.
+        return deform_windows(frame, x_grid, y_grid, u, -v)
```
antonioterpin: The `-v` flip lives at the call site, not in `deform_windows`'s API, which is a footgun for any caller that uses `deform_windows` directly (e.g. for custom multipass schemes or grid-refinement variants). The function's signature accepts `(u, v)` and silently applies `(x + u, y - v)` — i.e. `v` is in openpiv's image-y-positive-down convention. A user with a JAX/numpy-native flow field where `v` is y-positive-up will deform in the wrong direction and only notice via degraded multipass convergence. Either (a) take an explicit `image_convention: bool` arg, (b) rename the second arg to `neg_v` / document the convention contract in the docstring with an example, or (c) make `deform_windows` private (`_deform_windows`) and expose only `multipass_deform`.

src/flowgym/flow/open_piv/process.py
```diff
@@ -584,3 +584,173 @@ def upsample_flow(
...
+    )
+
+    # Resample the image onto the deformed grid.
+    pixel_x, pixel_y = jnp.meshgrid(side_x, side_y)
+    return jax.scipy.ndimage.map_coordinates(
+        frame, [pixel_y - vt, pixel_x + ut], order=1, mode="nearest"
+    )
+
+
+def multipass_deform(
```
antonioterpin, line 923: `multipass_deform` requires `window_size`, `search_area_size`, `overlap`, and `n_passes` to be Python ints (the `for _ in range(n_passes - 1)` and the `get_field_shape` call both bake them in at trace time), but the docstring never says so. Callers who naively `jax.jit(multipass_deform)` will either retrace every call or hit a tracer error on the Python `for`. The test `test_multipass_jit` correctly passes all four as `static_argnames` — please mirror that in the docstring (or, better, decorate the function with `@functools.partial(jax.jit, static_argnames=(...))` directly so callers get a JITed version by default).

tests/unit/flow/test_openpiv_jax_deform.py
```diff
@@ -0,0 +1,298 @@
...
+        )
+
+    a, b = jnp.asarray(img1)[None], jnp.asarray(img2)[None]
+    kwargs = {"window_size": ws, "search_area_size": sas, "overlap": ov}
+    err1 = endpoint_err(
+        np.asarray(multipass_deform(a, b, n_passes=1, **kwargs))[0]
+    )
+    err3 = endpoint_err(
+        np.asarray(multipass_deform(a, b, n_passes=3, **kwargs))[0]
+    )
```
antonioterpin: Weak monotonicity claim. The flow has ~2 px amplitude (`2.0 * sin(...)` + `1.5 * cos(...)`) and the threshold `err3 <= err1 + 1e-3` permits a 1e-3 px regression — at a typical subpixel error floor of ~0.05 px that is a 2% tolerance for regression. The whole point of window deformation is that error strictly decreases for smooth non-uniform flows; a tighter `err3 < 0.5 * err1` (or `err3 <= 0.9 * err1`) would actually defend against a future change that silently breaks the residual accumulation. Right now this test catches only catastrophic regressions.

src/flowgym/flow/open_piv/process.py
```diff
@@ -584,3 +584,173 @@ def upsample_flow(
...
+
+    - the window-centre field ``(u, v)`` is bilinearly interpolated onto the
+      pixel grid; outside the window-centre grid the values are clamped to
+      the border, exactly reproducing ``RectBivariateSpline`` with degree 1
+      (which extrapolates as a constant there); and
+    - the image is resampled at ``(y - vt, x + ut)`` with the same
+      ``map_coordinates`` linear interpolation and ``"nearest"`` border mode
+      as the reference.
+
+    Only the linear case is reproduced: openpiv's default cubic field
```
antonioterpin: The documented limitation 'openpiv's default cubic field interpolation (`RectBivariateSpline` degree 3) has no exact JAX equivalent' is the kind of thing that will be silently forgotten in 6 months when someone compares against openpiv with default settings and is confused by a ~5% subpixel discrepancy on sharp gradients. Please open a follow-up issue (cubic field interpolation via custom JAX kernel or `jax.scipy.signal.bspline`) and link it here as a `Note:` in the docstring, so the gap is tracked rather than buried in prose.

src/flowgym/flow/open_piv/process.py
```diff
@@ -584,3 +584,173 @@ def upsample_flow(
...
+    xs, ys = get_rect_coordinates((height, width), search_tuple, overlap_tuple)
+    x_grid = xs.reshape(n_rows, n_cols)
+    y_grid = ys.reshape(n_rows, n_cols)
+
+    def deform_batch(frame, u, v):
+        # "second image" method: deform with -v so frame_b is sampled at
+        # (y + v, x + u), the current displacement estimate.
+        return deform_windows(frame, x_grid, y_grid, u, -v)
+
+    flow = jnp.nan_to_num(flow)
```
antonioterpin, line 1008: Drift / outlier amplification across passes is unguarded. `jnp.nan_to_num(flow)` zeroes failed windows before deformation, then `extended_search_area_piv` computes a residual against a frame deformed by that artificial zero — in a region where the true displacement is, say, 5 px, the residual will be ~5 px and the accumulator becomes 0 + 5 = 5, which is correct on average but only because the residual happens to catch up. For a window where the *neighbor* failed (so the upsampled field has a ~zero hole in the middle of a non-zero region), `deform_windows` will linearly interpolate that hole and the next pass's residual won't fully recover it. openpiv's real recipe replaces outliers (`replace_outliers`) between passes precisely to avoid this. The docstring punts on this ('left to the caller') but there's no hook for the caller to inject replacement between passes — they would have to reimplement the loop. Consider taking an optional `between_passes: Callable | None = None` callback that receives `(flow, pass_index)` and returns a cleaned flow.

src/flowgym/flow/open_piv/process.py
antonioterpin: In general, please enforce agents to avoid this kind of changes documentation in docstrings...

src/flowgym/flow/open_piv/process.py
antonioterpin: In general, let's double check consistency on how we handle these things.

antonioterpin (changes requested):

src/flowgym/flow/open_piv/process.py
fbanelli: Documented (option c) in 24a0518, because changing the behaviour would break parity: the openpiv-primitive reference loop in `test_multipass_matches_openpiv_primitive_loop` itself `nan_to_num`s per pass and never re-applies NaN, so `n_passes>=2` zeroing failed windows is exactly what matches openpiv. Options (a)/(b) would diverge from that reference. The docstring now states the deliberate split loudly: `n_passes==1` keeps NaN (matching `extended_search_area_piv`); `n_passes>=2` zeroes via `nan_to_num`. The drift consequence you flag is real and tracked in #65.

src/flowgym/flow/open_piv/process.py
fbanelli: Done in 24a0518 — added the static-argname Note to `multipass_deform`'s docstring (mirroring `extended_search_area_piv`): `window_size`/`search_area_size`/`overlap`/`n_passes` are baked in at trace time and must be marked static under `jax.jit`. I kept it as a documented contract rather than decorating with `@functools.partial(jax.jit, ...)`, to stay consistent with `extended_search_area_piv` (which is also left un-jitted for the caller to wrap).

tests/unit/flow/test_openpiv_jax_deform.py
fbanelli: Tightened in 24a0518, but the data corrected the premise: error is **not** monotonically decreasing here. Measured median endpoint error is err1=0.206, err2=0.134, err3=0.179, err4=0.215 — it peaks at pass 2 then drifts up (pass 4 worse than pass 1). So `err3 < 0.5*err1` would fail and `err3 <= 0.9*err1` passes by only ~4% (fragile). Instead the test now asserts the real improvement where the method earns its keep — `err2 <= 0.8*err1` (~19% margin) — plus a no-worse guard `err3 <= err1`. The non-monotonic drift is tracked in #65.

src/flowgym/flow/open_piv/process.py
fbanelli: Opened #64 to track cubic field interpolation (`RectBivariateSpline` degree 3). Per the maintainer note on docstring prose, I trimmed the limitation paragraph in `deform_windows` to a single line that links #64 rather than expanding it.

src/flowgym/flow/open_piv/process.py
fbanelli: Good call — this is real, not theoretical. Measured: error peaks at pass 2 (0.134) and drifts back up by pass 4 (0.215 > single-pass 0.206), exactly the amplification you describe. Opened #65 with the evidence and your proposed `between_passes: Callable[[flow, pass_index], flow]` hook (default `None` = current behaviour). I held off adding the API here because it's part of the multipass-internals design the maintainer wants to review holistically (see the `-v` thread); #65 captures the proposal so it's decided deliberately.

src/flowgym/flow/open_piv/process.py
fbanelli: Deferring the API change here per your "double check consistency" note below — the `-v`/image-y-down convention, the NaN handling, and the between-pass outlier hook (#65) are all multipass-internals that are best decided together rather than piecemeal in this PR. Flagging for your consistency pass; happy to implement whichever shape you prefer (explicit `image_convention` arg, rename to `neg_v`, or making `deform_windows` private and exposing only `multipass_deform`).

src/flowgym/flow/open_piv/process.py
fbanelli: Acknowledged. Trimmed the cubic-interpolation limitation prose in `deform_windows` to a one-line link to #64. I kept the NaN-semantics and jit static-arg notes since those are behavioural/usage contracts the reviewers asked to pin (not limitation prose), but happy to trim further if you'd rather those live elsewhere.

tests/unit/flow/test_openpiv_jax_deform.py
```diff
@@ -0,0 +1,310 @@
...
+            "window_size",
+            "search_area_size",
+            "overlap",
+            "n_passes",
+        ),
+    )
+    eager = multipass_deform(
+        a, b, window_size=32, search_area_size=32, overlap=16, n_passes=3
+    )
+    compiled = jitted(
```
antonioterpin, line 302: The no-worse guard is fragile: measured EPE is 0.206→0.134→0.179→0.215→0.253, so this holds at pass 3 (0.179 ≤ 0.206) but would FAIL at pass 4 (0.215 > 0.206). The test passes only because it stops at `n_passes=3` — it doesn't actually defend against the drift it documents. Either assert the drift bound at the worst pass you intend to support, or gate/cap `n_passes>2`.

src/flowgym/flow/open_piv/process.py
```diff
@@ -584,3 +584,183 @@ def upsample_flow(
...
+    displacement, deform the second image toward the first by the current
+    estimate, re-correlate to obtain the residual, and accumulate. This is
+    the main accuracy lever for non-uniform flows that a single pass
+    under-resolves. The grid resolution (window/search/overlap) is held
+    fixed across passes, matching openpiv's ``deformation_method="second
+    image"`` recipe (the second image is deformed with ``-v`` so the sample
+    grid is ``(y + v, x + u)``).
+
+    Failed (NaN) windows are handled differently by pass count, by design:
+    ``n_passes == 1`` returns the single-pass field with NaNs intact (matching
```
antonioterpin, line 944: Drift confirmed empirically: without between-pass `replace_outliers` (#65), EPE peaks at pass 2 then degrades *below* the single-pass baseline by pass 4. For a feature billed as 'the main accuracy lever', `n_passes>2` is actively harmful. I'd land the #65 hook here, or hard-cap/warn on `n_passes>2`, rather than only documenting it.

src/flowgym/flow/open_piv/process.py
```diff
@@ -584,3 +584,183 @@ def upsample_flow(
...
+    """
+    if DEBUG:
+        assert frame.ndim == 2, (
+            f"Frame must be 2D (height, width), instead {frame.shape}"
+        )
+        assert x.ndim == 2 and y.ndim == 2, (
+            "x and y must be 2D window-centre meshgrids."
+        )
+        assert u.shape == x.shape and v.shape == x.shape, (
+            "u and v must match the window-centre grid shape."
```
antonioterpin, line 887: With DEBUG off (the default) a 1×N or N×1 grid no longer raises: JAX clamps OOB indexing so `y1[1]→y1[0]`, `dy=0`, and the deformed image comes back silently non-finite (NaN/inf) rather than erroring. The assert only fires under DEBUG. Consider a cheap always-on shape check, or compute spacing in a way that fails loudly in production mode.

src/flowgym/flow/open_piv/process.py
```diff
@@ -584,3 +584,183 @@ def upsample_flow(
...
+    # Resample the image onto the deformed grid.
+    pixel_x, pixel_y = jnp.meshgrid(side_x, side_y)
+    return jax.scipy.ndimage.map_coordinates(
+        frame, [pixel_y - vt, pixel_x + ut], order=1, mode="nearest"
+    )
+
+
+def multipass_deform(
+    img1: jnp.ndarray,
+    img2: jnp.ndarray,
```
antonioterpin, line 925: The `-v` image-y-down convention lives only at this call site; direct `deform_windows` callers get no API signal and will silently mis-orient. The sign is numerically correct (matches openpiv 'second image'), but per the first-pass discussion this footgun is worth resolving before merge — explicit `image_convention` arg, `neg_v` rename, or making `deform_windows` private and exposing only `multipass_deform`.

antonioterpin (changes requested):
**Second pass — REQUEST CHANGES: two design issues I'd resolve here, not defer.**

The numerical parity is solid: `deform_windows` reproduces openpiv's linear `create_deformation_field` + `map_coordinates` to ~1e-4, the `-v` sign matches `deformation_method="second image"`, all 13 tests pass.

**1. The drift is worse than 'non-monotonic'.** Measured EPE 0.206 → 0.134 → 0.179 → 0.215 → 0.253 — it degrades *below* the single-pass baseline by pass 4. For a feature billed as 'the main accuracy lever', more passes actively hurt; only `n_passes==2` is safe without between-pass outlier replacement. I'd land the #65 `between_passes`/`replace_outliers` hook **in this PR**, or hard-cap/refuse `n_passes>2`, rather than ship a footgun whose recommended setting (more passes) makes results worse. The no-worse test guard is also fragile (it only passes because it stops at pass 3 — see inline).

**2. `-v` convention + NaN asymmetry.** The `-v` image-y-down convention lives only at the call site (footgun for direct `deform_windows` callers), and the NaN handling flips at the 1→2 pass boundary. Per my earlier note these multipass-internals are worth resolving here (explicit `image_convention` arg / `neg_v` rename / make `deform_windows` private) — happy to pick the shape, just don't want them bounced to follow-ups piecemeal.

Minor inline: with DEBUG off a 1×N/N×1 grid now yields a silently non-finite image (JAX clamps OOB indexing) rather than erroring.

src/flowgym/flow/open_piv/process.py
antonioterpin: Fixed in 1a79f3b: the >= 2-points spacing check is now an **always-on** `ValueError` in `_deform_windows` (the grid shape is static, so it fails at trace time regardless of `DEBUG`). A 1xN/Nx1 grid no longer silently returns a non-finite image. Pinned by `test_deform_windows_grid_guard` (asserts it raises with `DEBUG` off, the default).

src/flowgym/flow/open_piv/process.py
antonioterpin: Fixed in 1a79f3b: made `deform_windows` private (`_deform_windows`) and kept `multipass_deform` as the only public entry point — it owns the `-v` sign. No public API carries the image-y-down convention now (option (c) from the first-pass thread). The docstring also states the convention for future maintainers.

src/flowgym/flow/open_piv/process.py
antonioterpin: Landed the #65 hook in 1a79f3b: `multipass_deform(..., between_passes: Callable[[flow, pass_index], flow] | None = None)`. Callers can inject outlier replacement / smoothing between passes; default `None` keeps openpiv parity. One caveat worth flagging: on the smooth synthetic test the drift is **accumulation overshoot, not outlier-driven** (a median/`replace_outliers` cleaner actually over-smooths and hurts there), so the hook is the general mechanism and the right cleaner is data-dependent. Wiring (called per pass, output threaded into the next pass) is pinned by `test_multipass_between_passes_hook`.

tests/unit/flow/test_openpiv_jax_deform.py
antonioterpin: Fixed in 1a79f3b: dropped the fragile `err3 <= err1` no-worse guard. `test_multipass_improves_accuracy` now asserts only the real win at the supported pass (`err2 <= 0.8*err1`); the default path is not claimed beyond pass 2 (it drifts — documented + #65), so there's no untested drift claim. Higher passes are supported via the `between_passes` hook, exercised in `test_multipass_between_passes_hook`.

antonioterpin:
-

PR #61, opened 2026-06-06
feat(open_piv): port sig2noise_val outlier mask

fbanelli:
Branches off `dev`, independent of the other open PRs.

## What

Adds `sig2noise_val`, the JAX port of openpiv's `validation.sig2noise_val`: flag vectors whose signal-to-noise ratio is below a threshold. This **completes the signal-to-noise outlier-rejection path** — `sig2noise_ratio` (ported in #55) computes the ratio, and `sig2noise_val` turns it into a boolean outlier mask. Previously flowgym computed the ratio in JAX but had no JAX step to threshold it (only an external openpiv call).

It's a one-liner — `s2n < threshold`, `True` = outlier — and NaN ratios stay unflagged (`nan < threshold` is `False`), matching the reference exactly.

## Tests (1-to-1 openpiv parity, in the established files)

- `test_openpiv_jax_numerical.py`: parity vs `validation.sig2noise_val` on 2D and 3D arrays across thresholds; NaN-not-flagged parity; and on **real pipeline ratios** — this last one feeds the JAX pipeline's s2n to *both* implementations, isolating the threshold step from the documented `sig2noise_ratio` flag divergence (see #55).
- `test_openpiv_jax_jit.py`: jit-equivalence.

101 openpiv tests pass locally (CPU). ruff (pinned 0.14.1), pydoclint, basedpyright clean.

## Note

This returns openpiv's convention (`True` = outlier), matching the reference for parity. Wiring it into flowgym's postprocess validation framework (which uses `True` = valid and `(flow, mask, state)` tuples) would be a thin follow-up wrapper if desired.

### DIFF
--- a/src/flowgym/flow/open_piv/process.py
+++ b/src/flowgym/flow/open_piv/process.py
@@ -372,12 +372,21 @@ def sig2noise_val(s2n: jnp.ndarray, threshold: float = 1.0) -> jnp.ndarray:
     NaN entries in ``s2n`` yield ``False`` (not flagged), since ``nan <
     threshold`` is ``False``, matching the reference.
 
+    Note:
+        This returns openpiv's convention, where ``True`` marks an *outlier*.
+        That is the **inverse** of flowgym's ``postprocess`` mask convention
+        (``True``/``1`` means *valid*; see ``postprocess.data_validation``).
+        Invert the result (``~mask``) before composing it with ``postprocess``
+        masks, otherwise it rejects exactly the vectors it should keep. The
+        openpiv name is kept for port parity.
+
     Args:
         s2n: Signal-to-noise ratios of any shape.
         threshold: Vectors with ``s2n < threshold`` are flagged.
 
     Returns:
-        Boolean array of the same shape as ``s2n``; ``True`` marks outliers.
+        Boolean array of the same shape as ``s2n``; ``True`` marks outliers
+        (openpiv convention -- see the Note above).
     """
     return s2n < threshold
 
--- a/src/flowgym/flow/open_piv/process.py
+++ b/src/flowgym/flow/open_piv/process.py
@@ -361,6 +361,27 @@ def sig2noise_ratio(
     return jnp.where(flag, 0.0, ratio)
 
 
+def sig2noise_val(s2n: jnp.ndarray, threshold: float = 1.0) -> jnp.ndarray:
+    """Flag vectors whose signal-to-noise ratio is below a threshold.
+
+    JAX port of ``openpiv.validation.sig2noise_val``: vectors whose
+    signal-to-noise ratio (e.g. from :func:`sig2noise_ratio`) is below
+    ``threshold`` are marked as outliers. This completes the standard
+    signal-to-noise outlier-rejection path.
+
+    NaN entries in ``s2n`` yield ``False`` (not flagged), since ``nan <
+    threshold`` is ``False``, matching the reference.
+
+    Args:
+        s2n: Signal-to-noise ratios of any shape.
+        threshold: Vectors with ``s2n < threshold`` are flagged.
+
+    Returns:
+        Boolean array of the same shape as ``s2n``; ``True`` marks outliers.
+    """
+    return s2n < threshold
+
+
 def subpixel_displacement(
     corr: jnp.ndarray,
     peaks_i: jnp.ndarray,
--- a/tests/unit/flow/test_openpiv_jax_jit.py
+++ b/tests/unit/flow/test_openpiv_jax_jit.py
@@ -24,6 +24,7 @@ from flowgym.flow.open_piv.process import (
     get_rect_coordinates,
     normalize_intensity,
     sig2noise_ratio,
+    sig2noise_val,
     sliding_window_array,
     subpixel_displacement,
     upsample_flow,
@@ -225,6 +226,16 @@ def test_sig2noise_ratio_jit(windows, sig2noise_method):
     _assert_same(
         sig2noise_ratio(corr, sig2noise_method=sig2noise_method, width=2),
         jitted(corr, sig2noise_method=sig2noise_method, width=2),
+    )
+
+
+def test_sig2noise_val_jit(windows):
+    """sig2noise_val traces under jit and matches its eager output."""
+    s2n = sig2noise_ratio(fft_correlate_images(windows, windows))
+    jitted = jax.jit(sig2noise_val, static_argnames="threshold")
+    np.testing.assert_array_equal(
+        np.asarray(sig2noise_val(s2n, threshold=1.0)),
+        np.asarray(jitted(s2n, threshold=1.0)),
     )
 
 
--- a/tests/unit/flow/test_openpiv_jax_numerical.py
+++ b/tests/unit/flow/test_openpiv_jax_numerical.py
@@ -32,7 +32,7 @@ vectorized reference exactly, flagged windows must be zero.
 import jax.numpy as jnp
 import numpy as np
 import pytest
-from openpiv import filters, pyprocess
+from openpiv import filters, pyprocess, validation
 
 from flowgym.flow.open_piv.openpiv_jax import replace_outliers
 from flowgym.flow.open_piv.process import (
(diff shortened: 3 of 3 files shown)


fbanelli:
Copilot generated no comments on this PR and I see nothing to change — `sig2noise_val` is a one-line thresholding step with parity, NaN-handling, and jit tests. Ready for review.

src/flowgym/flow/open_piv/process.py
```diff
@@ -361,6 +361,27 @@ def sig2noise_ratio(
...
+    Args:
+        s2n: Signal-to-noise ratios of any shape.
+        threshold: Vectors with ``s2n < threshold`` are flagged.
+
+    Returns:
+        Boolean array of the same shape as ``s2n``; ``True`` marks outliers.
+    """
+    return s2n < threshold
+
+
```
antonioterpin, line 393: The True=outlier convention here is the inverse of what every other mask in `flowgym/flow/postprocess/` uses (`1 means valid`, per `data_validation.py`, `data_smoothing.py`, `__init__.py`). Since this function lives one import away from that code, the inversion is a real foot-gun: a user composing `sig2noise_val(...)` with `postprocess.data_validation` will silently invalidate everything that should be kept.

The PR description acknowledges this and defers a wrapper, which is fine, but please make the divergence loud at the call site rather than only in the parity-focused note. Two concrete options:

1. Rename to `sig2noise_outlier_mask` (or `is_sig2noise_outlier`) so the inversion is in the name itself, and keep `sig2noise_val` only as a deprecated alias if openpiv-name parity matters.
2. Add a `Warning:` / `Note:` block at the top of the docstring (before Args) along the lines of: 'Returns openpiv's convention (True = outlier). This is the inverse of flowgym's postprocess mask convention (True/1 = valid); invert before passing to `postprocess.data_validation` etc.'

Option 1 is preferable — naming carries further than docstrings.

tests/unit/flow/test_openpiv_jax_jit.py
```diff
@@ -228,6 +229,16 @@ def test_sig2noise_ratio_jit(windows, sig2noise_method):
     )
 
 
+def test_sig2noise_val_jit(windows):
+    """sig2noise_val traces under jit and matches its eager output."""
+    s2n = sig2noise_ratio(fft_correlate_images(windows, windows))
+    jitted = jax.jit(sig2noise_val, static_argnames="threshold")
+    np.testing.assert_array_equal(
+        np.asarray(sig2noise_val(s2n, threshold=1.0)),
+        np.asarray(jitted(s2n, threshold=1.0)),
```
antonioterpin, line 238: This asserts `sig2noise_val(...) == jax.jit(sig2noise_val)(...)`, i.e. jitted vs eager of the *same* JAX function. That only proves the function traces — both sides will agree even if the JAX implementation drifts from openpiv under jit (e.g. dtype promotion, weak-type quirks around the Python float `threshold`). The numerical test in `test_openpiv_jax_numerical.py` compares against `validation.sig2noise_val` but is not jitted.

Consider either (a) jitting inside the numerical parity tests as well, or (b) making this test compare `jax.jit(sig2noise_val)(...)` directly against `validation.sig2noise_val(...)` on the same input. That closes the loop the other ported functions in this file already close.

src/flowgym/flow/open_piv/process.py
```diff
@@ -361,6 +361,27 @@ def sig2noise_ratio(
...
+    NaN entries in ``s2n`` yield ``False`` (not flagged), since ``nan <
+    threshold`` is ``False``, matching the reference.
+
+    Args:
+        s2n: Signal-to-noise ratios of any shape.
+        threshold: Vectors with ``s2n < threshold`` are flagged.
+
+    Returns:
+        Boolean array of the same shape as ``s2n``; ``True`` marks outliers.
+    """
```
antonioterpin, line 390: Worth documenting the dtype contract explicitly: `s2n < threshold` returns a JAX boolean array, but downstream `jnp.logical_*` consumers sometimes get tripped up by weak-typing in JAX. The numerical test asserts `got.dtype == bool` on the converted numpy side; consider an explicit `(s2n < threshold).astype(jnp.bool_)` in the implementation, or at minimum a one-line guarantee in the Returns section that the dtype is `jnp.bool_` regardless of the dtype of `s2n`.

antonioterpin: Let's open an issue to strictify typing (see https://github.com/antonioterpin/pinet)

antonioterpin (changes requested):

src/flowgym/flow/open_piv/process.py
fbanelli: Opened #63 to track strictifying the typing (pinet-style), and explicitly included the `sig2noise_val -> jnp.bool_` return-dtype contract there so it's pinned as part of that pass rather than ad hoc here.

tests/unit/flow/test_openpiv_jax_jit.py
fbanelli: Fixed in d8498ab via your option (a). `test_openpiv_jax_jit.py` is not openpiv-gated (collected without openpiv), so importing the reference there would break collection; instead I closed the loop in the gated numerical test: `test_sig2noise_val_matches_reference` now also runs `jax.jit(sig2noise_val)` and asserts the jitted output equals `openpiv.validation.sig2noise_val`, not just the eager JAX result. The jit-file test stays as a pure tracing/eager-equivalence smoke check.

src/flowgym/flow/open_piv/process.py
fbanelli: Addressed via option 2 in d8498ab — added a prominent `Note:` at the top of the docstring: this returns openpiv's `True == outlier` convention, the **inverse** of flowgym's postprocess `True/1 == valid`, with an explicit instruction to invert (`~mask`) before composing with `postprocess.data_validation`. I kept the `sig2noise_val` name rather than renaming, because the whole port deliberately mirrors openpiv's names (fft_correlate_images, subpixel_displacement, ...) and a rename would break that convention. If you'd prefer option 1 (rename to `sig2noise_outlier_mask` + deprecated alias), I've noted it on the typing/naming follow-up #63 so it can be decided consistently across the port.

src/flowgym/flow/open_piv/process.py
fbanelli: Per your note below, tracked in #63 rather than patched ad hoc. For the record the dtype is already `jnp.bool_` today (`s2n < threshold` always yields a boolean array, and `test_sig2noise_val_matches_reference` asserts `got.dtype == bool`), so this is a "pin the guarantee" item — which #63 will do as part of the strict-typing pass (incl. an explicit return-type annotation).

src/flowgym/flow/open_piv/process.py
```diff
@@ -361,6 +361,36 @@ def sig2noise_ratio(
     return jnp.where(flag, 0.0, ratio)
 
 
+def sig2noise_val(s2n: jnp.ndarray, threshold: float = 1.0) -> jnp.ndarray:
```
antonioterpin, line 364: The docstring Note resolves my first-pass concern. For the #63 rename decision: I'd prefer renaming to `sig2noise_outlier_mask` now and keeping `sig2noise_val` as a deprecated alias. The inverse convention is invisible at the call site, and every mask in `flow/postprocess/` uses the opposite `1=valid` convention, so a self-documenting name removes the foot-gun where it actually bites (the compose boundary). Non-blocking.

antonioterpin (approved):
**Second pass — LGTM (one non-blocking naming suggestion inline).**

All three first-pass fixes landed (d8498ab): the inverse-convention Note is prominent (right before Args, with the explicit `~mask`-before-composing instruction), the parity test now runs both eager and `jax.jit(sig2noise_val)` against `openpiv.validation.sig2noise_val` across 2D/3D shapes with dedicated NaN-not-flagged and real-pipeline-ratio coverage, and `s2n < threshold` is an exact openpiv match (NaN → False, bool dtype). Dtype-pin deferred to #63 is fine.

On the rename: I lean toward doing it now (option 1) since the True=outlier convention is invisible at the call site and every mask in `flow/postprocess/` uses the opposite `1=valid` — but it's a non-blocking ergonomics improvement on correct, well-tested code.

PR #62, opened 2026-06-08
feat(scripts): generic, algorithm-agnostic cache collector + ensemble selector

fbanelli:
## Summary

Cache collection is already algorithm-agnostic at the framework level —
`src/main.py --mode eval` dispatches through the polymorphic `enrich()` and
`get_cache_id_suffix()`, so DIS/RAFT/openpiv/art_of_piv all flow through one
path. This adds a clean, estimator-agnostic way to fill and consume caches,
**without** a per-algorithm producer script.

- **`src/main.py`**: add `--cache-root` / `--cache-id` flags. A new helper
  `apply_cache_cli_overrides` (`run_setup.py`) injects them into the dataset's
  caching block, so the dataset YAML carries only the `spec` (the one
  genuinely estimator-specific bit) and the same config can fill many caches
  without edits.
- **`scripts/collect_cache.py`** (new): thin sweep wrapper that runs the eval
  once per estimator config into a shared `--cache-root` (subprocess-isolated
  so JIT/GPU memory is released between configs; continue-on-error; end
  summary). No algorithm-specific logic.
- **`scripts/select_ensemble.py`** (new): picks a size-`K` subset minimizing
  the aggregate per-image best error (greedy, plus an exact MILP for the mean
  aggregator with `--exact`). `--metric` selects the parquet error column
  (default `epe`); `timing.json` is optional (timing is device-dependent and
  collected separately from the device-independent error).
- **`docs/examples/caching.md`**: documents the three commands.

## Usage

```bash
# one cache (no YAML edits — spec stays in the dataset config)
uv run python src/main.py --mode eval --estimator e.yaml --dataset d.yaml --cache-root caches/

# sweep many configs into one root
uv run python scripts/collect_cache.py --models estimators/*.yaml --dataset d.yaml --cache-root caches/

# pick a K-subset from the cache
uv run python scripts/select_ensemble.py --cache-root caches/ --K 3 --metric epe
```

## Test plan

- [x] Unit tests (13): `apply_cache_cli_overrides`, `collect_cache.build_eval_command`
      + loop return codes, `select_ensemble` over a tiny on-disk cache (custom
      metric, with/without `timing.json`).
- [x] `ruff`, `ruff format`, `pydoclint`, `basedpyright`, push-stage hooks pass.

## Note

This is the public counterpart of the internal version of the same feature.
The selector is new here (the internal repo's earlier DIS-specific selector
was never mirrored to this fork), so it lands directly in its generalized form.

### DIFF
--- a/scripts/collect_cache.py
+++ b/scripts/collect_cache.py
@@ -65,7 +65,9 @@ def materialize_model_configs(
 
     Each entry is emitted as a ``{estimator, estimate_type, config}`` YAML
     (the ``name`` key, which is list-only metadata, is dropped) so it can be
-    passed to ``src/main.py --model``.
+    passed to ``src/main.py --model``. The entry ``idx`` prefixes every
+    filename so two entries that share a ``name`` (or sanitize to the same
+    stem) never overwrite each other.
 
     Args:
         entries: Estimator-entry dicts from :func:`load_estimators_list`.
@@ -79,7 +81,7 @@ def materialize_model_configs(
         model = {k: v for k, v in entry.items() if k != "name"}
         raw_stem = str(entry.get("name", f"model_{idx}"))
         stem = "".join(c if c.isalnum() or c in "-_" else "_" for c in raw_stem)
-        out = workdir / f"{stem}.yaml"
+        out = workdir / f"{idx:03d}_{stem}.yaml"
         out.write_text(yaml.dump(model), encoding="utf-8")
         paths.append(out)
     return paths
--- a/scripts/collect_cache.py
+++ b/scripts/collect_cache.py
@@ -25,9 +25,64 @@ from __future__ import annotations
 import argparse
 import subprocess
 import sys
+import tempfile
 from pathlib import Path
 
+import yaml
+
 DEFAULT_RUNNER = "uv run python"
+
+
+def load_estimators_list(path: Path) -> list[dict]:
+    """Load the ``estimators:`` entries from an estimators_list YAML.
+
+    This is the multi-estimator format consumed by ``art_of_piv`` (e.g.
+    ``dis_models_20.yaml``): a top-level ``estimators:`` key holding a list
+    of ``{name, estimator, estimate_type, config}`` dicts.
+
+    Args:
+        path: Path to the estimators_list YAML.
+
+    Returns:
+        The list of estimator-entry dicts.
+
+    Raises:
+        ValueError: If the file has no non-empty ``estimators:`` list.
+    """
+    with path.open("r", encoding="utf-8") as fh:
+        data = yaml.safe_load(fh)
+    estimators = data.get("estimators") if isinstance(data, dict) else None
+    if not estimators:
+        raise ValueError(f"{path} has no non-empty `estimators:` list")
+    return estimators
+
+
+def materialize_model_configs(
+    entries: list[dict],
+    workdir: Path,
+) -> list[Path]:
+    """Write estimators_list entries as standalone model config YAMLs.
+
+    Each entry is emitted as a ``{estimator, estimate_type, config}`` YAML
+    (the ``name`` key, which is list-only metadata, is dropped) so it can be
+    passed to ``src/main.py --model``.
+
+    Args:
+        entries: Estimator-entry dicts from :func:`load_estimators_list`.
+        workdir: Directory to write the per-entry YAMLs into.
+
+    Returns:
+        Paths to the written model config YAMLs, in entry order.
+    """
+    paths: list[Path] = []
+    for idx, entry in enumerate(entries):
+        model = {k: v for k, v in entry.items() if k != "name"}
+        raw_stem = str(entry.get("name", f"model_{idx}"))
+        stem = "".join(c if c.isalnum() or c in "-_" else "_" for c in raw_stem)
+        out = workdir / f"{stem}.yaml"
+        out.write_text(yaml.dump(model), encoding="utf-8")
+        paths.append(out)
+    return paths
 
 
 def build_eval_command(
@@ -82,8 +137,17 @@ def _parse_args(argv: list[str] | None = None) -> argparse.Namespace:
         "--models",
         type=Path,
         nargs="+",
-        required=True,
-        help="Model config YAMLs to evaluate (one cache each).",
+        default=None,
+        help="Standalone model config YAMLs to evaluate (one cache each).",
+    )
+    parser.add_argument(
+        "--estimators-list",
+        type=Path,
+        nargs="+",
+        default=None,
+        help="estimators_list YAMLs (a top-level `estimators:` list, e.g. "
+        "dis_models_20.yaml). Every sub-estimator is collected as its own "
+        "cache; nested lists dedupe by cache_id, so the superset suffices.",
     )
     parser.add_argument(
         "--dataset",
@@ -122,7 +186,10 @@ def _parse_args(argv: list[str] | None = None) -> argparse.Namespace:
         action="store_true",
         help="Stop at the first failing config instead of continuing.",
     )
-    return parser.parse_args(argv)
+    args = parser.parse_args(argv)
+    if not args.models and not args.estimators_list:
+        parser.error("provide --models and/or --estimators-list")
+    return args
 
 
 def main(argv: list[str] | None = None) -> int:
@@ -135,7 +202,16 @@ def main(argv: list[str] | None = None) -> int:
         Process exit code (0 if every config succeeded, 1 otherwise).
     """
     args = _parse_args(argv)
-    models = args.models if args.limit is None else args.models[: args.limit]
+
+    models: list[Path] = list(args.models or [])
+    if args.estimators_list:
+        workdir = Path(tempfile.mkdtemp(prefix="collect_cache_"))
+        for list_path in args.estimators_list:
+            entries = load_estimators_list(list_path)
+            models.extend(materialize_model_configs(entries, workdir))
+
+    if args.limit is not None:
+        models = models[: args.limit]
     runner = tuple(args.runner.split())
 
     n_ok = 0
--- a/scripts/collect_cache.py
+++ b/scripts/collect_cache.py
@@ -1,0 +1,169 @@
+"""Fill per-image error caches for many model configs over one dataset.
+
+Estimator-agnostic sweep producer. For each candidate model config it runs
+``src/main.py --mode eval`` with a shared ``--cache-root``; the dataset
+config supplies the caching ``spec`` (e.g. ``{epe: [float32, []]}``), and
+each estimator's ``get_cache_id_suffix`` keeps the per-config caches in
+separate ``<cache-root>/<cache_id>/`` directories. Nothing here is specific
+to any algorithm -- DIS, RAFT, openpiv, art_of_piv all flow through the same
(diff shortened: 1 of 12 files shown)


scripts/select_ensemble.py
```diff
@@ -0,0 +1,814 @@
...
+        return float(np.mean(values))
+    if aggregator == "median":
+        return float(np.median(values))
+    raise ValueError(f"unknown aggregator {aggregator!r}")
+
+
+def _greedy_select(
+    E: np.ndarray,
+    K: int,
+    aggregator: str,
```
antonioterpin, line 361: `scipy` is imported unconditionally at module top but isn't a declared dependency in `pyproject.toml` (absent from both core `dependencies` and the `dev` group) — it only resolves transitively via jax. This makes two PR claims false: (1) the script imports cleanly without scipy, and (2) `--exact` degrades gracefully when the solver is missing — in fact a missing scipy raises `ModuleNotFoundError` at import, killing even the greedy path and the test module. Either declare scipy as a dependency, or move the scipy import into `_exact_milp` and guard `--exact` with a try/except that falls back to greedy with a warning.

scripts/collect_cache.py
```diff
@@ -0,0 +1,245 @@
...
+        type=Path,
+        required=True,
+        help="Dataset config YAML; must carry a caching block with a spec.",
+    )
+    parser.add_argument(
+        "--cache-root",
+        type=Path,
+        required=True,
+        help="Shared output root for the per-config caches.",
+    )
```
antonioterpin, line 165: The output filename is derived solely from the (sanitized) entry `name` with no de-duplication. If two estimators-list entries share a `name` (or sanitize to the same stem), the second `out.write_text` overwrites the first and both returned paths point to the same file — one config is silently dropped and another evaluated twice. The `idx` fallback only applies when `name` is missing entirely. Fold `idx` into the stem (e.g. `f"{idx:03d}_{stem}.yaml"`) to guarantee uniqueness.

scripts/select_ensemble.py
```diff
@@ -0,0 +1,814 @@
...
+    per_image_winner_global = np.asarray(selected)[per_image_argmin_local]
+
+    win_counts = np.bincount(per_image_argmin_local, minlength=len(selected))
+
+    entries: list[dict[str, Any]] = []
+    for local_idx, a in enumerate(selected):
+        entries.append(
+            {
+                "cache_id": cache_ids[a],
+                "config": configs[a].get("config", {}),
```
antonioterpin, line 584: `config` here must come from `timing.json` — the in-repo cache writer (`_write_meta`) only emits `cache_id`/`version`/`spec`, never a `config` key, and nothing in this repo writes `timing.json`. So `--export-models` produces an empty/`skipped` export for any cache not accompanied by an externally-supplied `timing.json` carrying `config`. Worth a note in the help/docstring that `export-models` depends on an external `timing.json` schema, since it's not discoverable from the repo.

tests/unit/scripts/test_select_ensemble.py
```diff
@@ -0,0 +1,205 @@
...
+        ]
+    )
+    assert rc == 0
+    data = yaml.safe_load(out.read_text())
+    sizes = {e["config"]["patch_size"] for e in data["estimators"]}
+    assert sizes == {7, 9}
+    assert all(e["estimator"] == "dis_jax" for e in data["estimators"])
+
+
+def test_finite_time_limit_without_timing_drops_all(tiny_cache: Path) -> None:
```
antonioterpin, line 284: The 13 tests cover greedy correctness, overrides, loaders and export well, but the `--exact` MILP path (`_exact_milp` / `_build_milp_constraints`) has no test at all — neither that it returns a valid K-subset nor that it matches greedy on the tiny cache. Given the MILP is the more error-prone code (and the only place the currently-undeclared scipy dep is exercised end-to-end), add at least one `--exact` test asserting the exact selection equals the known optimum (and `gap≈0` vs greedy).

antonioterpin (changes requested):
**First pass — one blocking dependency issue, plus secondary robustness items.**

Nice work — the selector core is sound: the greedy correctly minimizes the mean/median over per-image best error, and the HiGHS/scipy MILP (assignment polytope `Σ_a y_{a,i}=1`, `y≤x`, `Σ x=K`) provably matches the greedy mean objective. `apply_cache_cli_overrides` correctly preserves `spec` and other fields (in-place mutation is consistent with the surrounding `run_setup.py`), and the DIS `preset` enum-name→int mapping is correct.

**Blocking:** `scipy` is imported unconditionally at module top but is **not a declared dependency** in `pyproject.toml` (neither core nor dev) — it only resolves transitively via jax today. This makes the PR's `--exact` 'graceful degradation' claim false: a missing scipy raises `ModuleNotFoundError` at import, killing even the greedy path and the whole test module. Declare scipy, or move the import into `_exact_milp` behind a try/except that falls back to greedy with a warning. (Also: `pyarrow`/`numpy` are dev-group only, so the scripts can't run in a plain install.)

Secondary items inline: filename collision in `collect_cache`, `export-models` depending on an external `timing.json` schema nothing in-repo writes, and the `--exact` MILP path having zero test coverage.

tests/unit/scripts/test_select_ensemble.py
antonioterpin: also would be good to test the flow: greedy as init + refinement with MILP

scripts/select_ensemble.py
```diff
@@ -0,0 +1,814 @@
...
+
+# DIS serializes its `preset` enum by name in get_config()/timing.json, but
+# the estimator constructor only accepts an int (or PresetType). Map known
+# names back to ints so exported configs re-load directly.
+_PRESET_NAME_TO_INT = {
+    "ULTRAFAST": 0,
+    "FAST": 1,
+    "MEDIUM": 2,
+    "HIGH_QUALITY": 3,
+}
```
antonioterpin: isnt this script supposed to be agnostic to DIS?

antonioterpin: (to the actual algorithm)

scripts/select_ensemble.py
```diff
@@ -0,0 +1,814 @@
...
+    parser.add_argument(
+        "--time-stat",
+        choices=("mean_ms", "median_ms", "min_ms", "max_ms"),
+        default="mean_ms",
+        help="Which timing.json field to compare against --time-limit.",
+    )
+    parser.add_argument(
+        "--exact",
+        action="store_true",
+        help="Run exact MILP after greedy (only supported for mean).",
```
antonioterpin: why only for mean? and not median?

scripts/select_ensemble.py
```diff
@@ -0,0 +1,814 @@
...
+    )
+    return E, cache_ids, records
+
+
+def _aggregate(values: np.ndarray, aggregator: str) -> float:
+    """Apply the configured aggregator to a 1-D EPE vector.
+
+    Args:
+        values: Per-image EPE values.
+        aggregator: ``"mean"`` or ``"median"``.
```
antonioterpin, line 343: also because here seems median is supported... I would make sure we are consistent in the docstrings

scripts/select_ensemble.py
antonioterpin: I think scipy should just be a core dependency

antonioterpin (changes requested):

scripts/select_ensemble.py
antonioterpin: Done in `1a50e9e` — `scipy>=1.11.0` is now a core dependency (added to `[project] dependencies` and recorded in `uv.lock`), so the module-top `from scipy.optimize import …` resolves in a plain install rather than only transitively via jax.

scripts/collect_cache.py
antonioterpin: Done in `1a50e9e` — `materialize_model_configs` now prefixes each emitted YAML with the entry index (`f"{idx:03d}_{stem}.yaml"`), so two entries sharing a `name` (or sanitizing to the same stem) get distinct files instead of one silently overwriting the other. Added `test_materialize_dedupes_duplicate_names`.

scripts/select_ensemble.py
antonioterpin: Done in `1a50e9e` — documented in both the `--export-models` help and the `_export_models` docstring that each candidate's `config` is read from its `timing.json` (an external schema; the in-repo cache writer emits only `meta.json` with `cache_id`/`version`/`spec`), and config-less candidates are skipped.

tests/unit/scripts/test_select_ensemble.py
antonioterpin: Done in `1a50e9e` — added two MILP tests: `test_exact_matches_known_optimum` asserts `--exact` returns the known optimum on the tiny cache with a zero greedy/exact gap, and `test_exact_milp_refines_greedy` builds a cache where greedy seeds a trap (mean 1.75) and the MILP refines it to the true optimum `{cand_b, cand_c}` (mean 1.0) — i.e. greedy-init → MILP-refinement, as suggested.

scripts/select_ensemble.py
antonioterpin: Good catch — fixed in `1a50e9e`. Removed the DIS-specific `_PRESET_NAME_TO_INT`/`_normalize_config` from the selector, which now emits configs verbatim and is fully algorithm-agnostic. The round-trip is closed in DIS itself instead: `DISJAXFlowFieldEstimator.__init__` now accepts the preset enum **name** (what `get_config()` emits), so exported configs re-load without the selector knowing anything about presets. Added a round-trip test in `test_dis_cache.py`.

scripts/select_ensemble.py
antonioterpin: Documented in `1a50e9e`. The MILP objective is the linear sum `Σ e_{a,i}·y_{a,i}` (= the mean up to the constant `1/M`); the median is a non-linear order statistic the assignment LP can't express, so exact is mean-only by construction and `--exact` is a no-op for `--aggregator median`. Spelled out in the module docstring, the `--exact` help, and the runtime note.

scripts/select_ensemble.py
antonioterpin: Done in `1a50e9e` — aligned the docs: greedy and `_aggregate` support both mean and median, while `--exact` is mean-only (the MILP minimizes the linear sum = mean; the median is a non-linear order statistic). The module docstring, `--exact` help, and runtime note now state this consistently.

antonioterpin (approved):
All review threads addressed in `1a50e9e`:
- `scipy` is now a core dependency (module-top import valid in a plain install);
- `collect_cache` prefixes materialized configs with their index so duplicate names no longer overwrite each other;
- `--export-models` documents that `config` comes from each candidate's external `timing.json`;
- the `--exact` MILP path is now tested (matches the known optimum at gap 0, and refines a greedy-suboptimal selection — greedy-init → MILP-refinement);
- the selector is fully algorithm-agnostic again: the DIS-specific preset normalization is gone, and the DIS constructor accepts its preset enum name so configs round-trip;
- mean-vs-median is documented consistently (exact is mean-only because the MILP objective is the linear sum = mean).

Quality gates pass on the changed files; 39 tests pass. LGTM.

albonomi02 (approved):
LGTM

Issue #68, opened 2026-06-09
Kinematic training for LIMA (random displacement-field generation) — blocked on synthpix


antonioterpin:
Follow-up to #67 / PR #66.

## Goal
Train LIMA (and other flow estimators) with **true kinematic training** in the sense of Manickathan, Mucignat & Lunati (*Meas. Sci. Technol.* 33, 124006, 2022; *Exp. Fluids* 64, 161, 2023): supervise on **randomly generated displacement fields**, not on stored CFD/DNS fields. This gives an effectively unlimited, cheap training distribution and is the strategy the LIMA papers actually use (18,278 random fields, no Navier–Stokes solves).

Per LIMA-1 Table 1, each field is a smooth random field: draw two i.i.d. white-noise component maps, low-pass each with a Gaussian of std `σ` (filter size **5–100 px**, drawn per field), then rescale so the field's max magnitude equals a per-field target `max(ds_ref)` drawn in **0–16 px**. Particle diameter (1–4 px), seeding density (0.05–0.40 ppp) and particle loss are already handled by synthpix's image renderer.

## Why it's blocked (synthpix)
synthpix can only source displacement fields **from disk**: every scheduler/data-source requires a non-empty `file_list` with a `.mat`/`.npy`/`.h5` extension, and the only generative component (`SyntheticImageSampler`) *consumes* fields and *emits images*. There is **no random displacement-field generator** and no non-file field source. PR #66 therefore only does the rendering half of kinematic training (randomized images per batch) on top of the **fixed, real** class-1 fields — it is supervised training on class-1, not kinematic training.

## What's needed (in synthpix, upstream)
A non-file, generative flow-field source that plugs into `synthpix.make()` so a dataset config can request kinematically-generated fields instead of a `file_list`, e.g.:
```yaml
scheduler_class: "kinematic"
flow_generator:
  image_shape: [256, 256]
  num_examples: 18278
  filter_sigma_range: [5, 100]
  max_displacement_range: [0, 16]
  seed: 0
```
Sketch: a `KinematicFlowFieldDataSource` (Grain `RandomAccessDataSource`, no `file_list`) whose `__getitem__` returns a freshly generated `(H, W, 2)` field via a JAX generator (Gaussian-filtered random noise, per-field `σ` and `max|ds|`, fold the record index into the PRNG key). It drops into the existing `grain.DataLoader`/`IndexSampler` path with no scheduler changes; `include_images: false` then renders fresh particle images on top — both halves now random.

[...]

Issue #67, opened 2026-06-09
LIMA PIV flow estimator + supervised training/eval on PIV class-1


antonioterpin:
Tracking issue for the work in PR #66.

## Summary
Adds **LIMA** (lightweight image matching architecture for PIV) as a JAX/Flax `FlowFieldEstimator` (`lima_piv`), re-implementing Manickathan, Mucignat & Lunati (*Exp. Fluids* 2023) with the padding/search-range improvements of Mucignat, Zdybał & Lunati (*Phys. Fluids* 37, 105112, 2025), plus a supervised training run and evaluation on the **PIV class-1** dataset.

## What this is
- LIMA-6 (LIMAR2): shared 6-level encoder (Table I) + weight-shared dilated decoder (Table III), symmetric warping, local correlation, iterative residual refinement; replicate padding, search range 2. ~0.93M params.
- **Supervised training on the PIV class-1 train split** (`activefluidcontrol/piv-class1`): the displacement fields are the dataset's *real* fields (DNS turbulence, JHTDB, SQG, backstep, cylinder, uniform), and synthpix renders randomized particle images on top of them each batch (`include_images: false`).
- Evaluation on the class-1 **test** split: best checkpoint **0.065 px mean EPE** (median 0.055), ~26× below the random-init baseline (1.72 px); ~0.15 px median on the real recorded images. Per-scenario breakdown in `experiments/lima/RESULTS.md`.

## What this is NOT
- This is **not** the random-displacement-field "kinematic training" of Manickathan et al. (*Meas. Sci. Technol.* 2022). The flow fields here come from the class-1 dataset (real CFD/analytic fields), not from a random-field generator. Only the particle-image *rendering* is randomized per batch.
- True kinematic training is tracked separately and is currently blocked on synthpix — see the companion issue.

## Notes
- A multi-agent review caught and fixed an inverted symmetric-warp sign; a directional warp test + end-to-end recovery test now pin the convention.
- Tabulated architecture is ~0.93M params vs the paper's stated 1.6M (the 2025 tables omit a separate flow-estimator stack).
- Reduced-budget run (~6% of the paper's 200-epoch budget); a late-training regression on the `uniform` scenario (lr 5e-4 without decay, best-checkpoint selection disabled) is documented — use the 32k checkpoint. Follow-ups: lr decay + validation-based `save_only_best`.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #69, opened 2026-06-09
feat(lima): true kinematic training via synthpix KinematicDataSource

antonioterpin:
## Summary

Adds **true kinematic training** for the LIMA PIV estimator — supervising on
*randomly generated* displacement fields (Manickathan et al. 2022), the strategy
that issue #68 tracked as **blocked on synthpix**. synthpix **0.3.1** ships an
in-memory `KinematicDataSource` with `scale_mode: "peak"` (paper-faithful
per-field peak-displacement normalisation), so the existing supervised driver
trains LIMA on purely random fields with **no estimator code changes**.

Closes #68. Builds on #66 / #67.

## What's new

- **synthpix lower bound:** `synthpix>=0.3.1` (and `synthpix[cuda12]>=0.3.1`);
  `uv.lock` regenerated.
- `experiments/lima/train_kinematic.yaml` — kinematic dataset config
  (`scheduler_class: kinematic`, no `file_list`, `include_images: false`), using
  **`scale_mode: peak`** with the paper defaults **sigma in [5,100], a in [0,16]**.
- `experiments/lima/lima_piv_kinematic.yaml` — estimator config (separate
  `out_dir`).
- `RESULTS.md` — Regime B (kinematic) results, refreshed for the peak config.

### Peak normalisation (replaces the earlier calibration hack)

synthpix generates each field as `ds = a*G_sigma*xi`. With `scale_mode: "peak"` each
field is normalised so its **peak displacement equals `a` px** — the paper's
"Maximum displacement, max(ds_ref)" (Manickathan, Mucignat & Lunati, *Exp.
Fluids* 64, 161, 2023, Table 1), and sigma-independent. So the paper ranges are used
directly. The earlier workaround (sigma in [5,30], a in [0,120], needed because synthpix
0.3.0 applied `a` as a raw multiplier that Gaussian filtering collapsed to
~0.10 px median) is **gone**.

## Results — class-1 test (LIMA-6 LIMAR2; re-run on the peak config, RTX 5090)

Trained **only on random fields** (never any class-1 field), evaluated on the
class-1 test split (EPE excl. 16 px border):

| Eval | Mean EPE | Median |
|---|---|---|
| random-init baseline | 1.717 px | 1.294 px |
| **best ckpt (32k), kinematic — rendered** | **0.094 px** | **0.073 px** |
| best ckpt, real recorded images | 0.254 px | 0.162 px |

An **~18×** reduction over baseline from purely random-field training, within
~1.5× of the in-distribution class-1-supervised model (#66: 0.065 px) and a
**better mean on real images** (vs 0.363 px). No late-training regression; the
`uniform`/translation case is now **0.038 px** (was 0.207 px under the linear
calibration). Full learning curve, per-scenario breakdown and A-vs-B head-to-head
in `experiments/lima/RESULTS.md`.

## Validation

- Re-ran 40k-batch training + class-1 eval (rendered + real) on an RTX 5090 with
  the peak config; the numbers above are from this run.
- 47 LIMA unit tests pass; pre-commit hooks (ruff, format, pydoclint, gitlint,
  check-yaml) green on the changed files.


experiments/lima/RESULTS.md
```diff
@@ -76,11 +96,116 @@ cylinder-near-wall real images (~0.5 px) where laser reflections and solid
 boundaries are hardest — the same regime where WIDIM also struggles (LIMA-1
 §3.3). The replicate padding (LIMAR) is specifically meant to help here.
 
+## Regime B — true kinematic training (random displacement fields)
+
+Resolves #68. Here the displacement fields themselves are randomly generated by
+synthpix's `KinematicDataSource` (new in synthpix 0.3.0): the model is trained
+**without ever seeing a class-1 field** and is then evaluated on the class-1
```
antonioterpin: lets not refer to the issue

antonioterpin: Done in a3581a5 — dropped the issue/PR references from RESULTS.md: both `Resolves #68.` lines are gone, and the Terminology note no longer cites `PR #66` / `issues #67/#68` (kept its substance: Regime A was mislabelled "kinematic", only the rendering was randomized there). No `#NN` references remain in the file.

experiments/lima/train_kinematic.yaml
```diff
@@ -0,0 +1,72 @@
...
+# filtered noise); it is NOT the paper's per-field peak-displacement target.
+# Gaussian filtering attenuates the noise amplitude strongly (and sigma-
+# dependently), so synthpix's own defaults (sigma 5-100, a 0-16) produce
+# degenerate sub-pixel fields (per-field peak |d| ~ 0.10 px median) -- useless
+# for training. The ranges below are CALIBRATED (see experiments/lima/RESULTS.md)
+# so per-field peak |d| spans a PIV-realistic ~0-18 px (median ~2.4 px),
+# overlapping the class-1 test distribution the model is evaluated on.
+scheduler_class: "kinematic"
+num_examples: 18278          # paper's random-field count (LIMA-1 Table 1)
+filter_sigma_range: [5, 30]  # Gaussian correlation length (px)
```
albonomi02: these valued must be changed in order to match the paper, scaling from 0 to 16 and sigma from 5 to 200

experiments/lima/train_kinematic.yaml
```diff
@@ -0,0 +1,72 @@
...
+scale_factor_range: [0, 120] # calibrated linear scale -> peak |d| ~0-18 px
+
+# Image generation parameters (randomized per batch -> data augmentation),
+# identical to train.yaml so the kinematic vs class-1 comparison is fair.
+image_shape: [256, 256]
+dt: 1
+seeding_density_range: [0.03, 0.04]
```
albonomi02, line 39: I'm not sure about this, but shouldn't we also match this values with the paper?

experiments/lima/RESULTS.md
```diff
@@ -1,19 +1,38 @@
...
+> only the *image rendering* was randomized there, not the displacement fields.
+> Regime B is kinematic training in the proper random-displacement-field sense;
+> the two are kept distinct below.
+
+## Regime A — supervised on class-1 fields (randomized rendering)
+
+### Setup
 
 - **Model:** LIMA-6, replicate padding, search range 2 (LIMAR2), the 2025
   paper's recommended config. ~0.93 M params (tabulated architecture).
```
albonomi02: let's cite correctly the paper from 2025

experiments/lima/RESULTS.md
```diff
@@ -76,11 +96,116 @@ cylinder-near-wall real images (~0.5 px) where laser reflections and solid
...
+### Field generator and calibration
+
+synthpix generates each field as `ds = a * G_sigma * xi` (white noise `xi ~ U(-1,1)`,
+Gaussian filter of width `sigma` px, per-field linear scale `a`; Manickathan et al.
+2022 §2.2). **Caveat:** synthpix applies `a` as a *linear* scale, not the
+paper's per-field peak-displacement target, and Gaussian filtering strongly
+(and sigma-dependently) attenuates amplitude. synthpix's own defaults (sigma in
+[5,100], a in [0,16]) therefore produce **degenerate sub-pixel fields** (per-field peak
+displacement ≈ 0.10 px median) — useless for training. We calibrate
+**sigma in [5,30], a in [0,120]** so the generated peaks span a PIV-realistic range
```
albonomi02: also here we should change the ranges to the ones of the paper

albonomi02:
we are not citing properly the papers in this PR, sometimes, we refer to the papers as the "202X one." let's fix this before merging

experiments/lima/train_kinematic.yaml
antonioterpin: Scaling is now `scale_factor_range: [0, 16]` (the peak/max displacement in px under `scale_mode: peak`). For the filter size I kept `[5, 100]`: both LIMA-0 (Table 3, "Filter size sigma 5-100") and LIMA-1 (Table 1, "Filter size sigma 5-100") give 5-100 for the kinematic-training Gaussian filter, and LIMA-2 reuses that same kinematic-training strategy without redefining the range.

The "1-200 pixels" in LIMA-2 is the wavelength of the synthetic sinusoidal shear flow used for the spatial-resolution test (u = A0 * sin(2*pi*y / wavelength)), measured at fixed seeding (0.2 ppp) and fixed particle diameter (2 px) on 500x2000 images -- it is not the training filter size. Happy to switch to 5-200 if you prefer, but 5-100 is what the kinematic-training tables specify.

experiments/lima/train_kinematic.yaml
antonioterpin: Good question -- I tried it. Matching the image-generator params to the paper (diameter 1-4 px, density 0.05-0.40 ppp, particle loss ~0-2%) and re-running the full 40k training + eval made the class-1 result ~15x worse: best checkpoint mean 1.14 / median 0.33 px (vs 0.094 / 0.073), and the final checkpoint diverged (uniform/translation EPE 9.67 px).

The cause is a train/eval rendering mismatch: the class-1 *test* images are rendered sparse (density 0.03-0.04 ppp, diameter 1.5-2.5), so the paper's dense distribution (up to 0.40 ppp) is out-of-distribution for this benchmark and also destabilizes the reduced-budget run at lr 5e-4. So for now the image-gen params stay matched to the class-1 eval rendering (train and eval agree); only the flow-field generator is paper-faithful (sigma 5-100, max displacement 0-16, peak). We could revisit by also rendering the eval at paper params, but then it is no longer the class-1 comparison.

experiments/lima/RESULTS.md
antonioterpin: Done (25bcb80) -- now cited as Mucignat, Zdybal & Lunati, Phys. Fluids 37, 105112 (2025), doi:10.1063/5.0283779, instead of "the 2025 paper".

experiments/lima/RESULTS.md
antonioterpin: Done -- this section now describes the peak-mode generator with the paper ranges (filter size 5-100, max displacement 0-16; LIMA-1 Table 1), and the old linear-calibration workaround is removed (25bcb80).

experiments/lima/RESULTS.md
antonioterpin: The last remaining issue reference ("Resolves issue #68" in train_kinematic.yaml) is removed in 25bcb80 -- no #NN references remain in the LIMA files now.

antonioterpin:
Citations fixed in 25bcb80: every loose "the 202X paper" / bare author-year is replaced with a named citation + DOI across RESULTS.md, train_kinematic.yaml, lima_piv.yaml, lima_piv_kinematic.yaml, eval_lima.py, flow/lima/lima_piv.py and nn/lima_model.py, all pointing at the RESULTS.md References section (LIMA-0/1/2 with DOIs). See the per-line replies on the specific threads.

PR #70, opened 2026-06-09
feat(scripts): generic per-config timing collector

fbanelli:
## What

Adds `scripts/collect_timing.py`, the **timing counterpart** to `collect_cache.py`. For a sweep of model configs (`--models` / `--estimators-list`, any estimator) it builds each estimator with `make_estimator` — exactly as `src/main.py --mode eval` does — and times the JIT-compiled forward, writing a `timing.json` into each config's cache dir.

Timing follows the repo's own `tests/test_dis_jax.py` speed tests:
- a random input pair of the dataset's `image_shape` is made **device-resident (on GPU when available) before timing** — PIV/optical-flow compute cost is data-independent, so no data loading is needed;
- the forward is JIT-compiled and warmed (compile cost recorded separately, excluded from the stats);
- measurement is `timeit.repeat(number=iters, repeat=repeat)`; the record reports `mean_ms`/`median_ms`/`min_ms`/`max_ms` (+ `std_ms`, `compile_ms`, batch size).

## Why a separate script

This is the **device-dependent** half deliberately split out of `collect_cache.py` (which stores only the **device-independent** per-image error, reusable across machines). `timing.json` is exactly what `select_ensemble.py --time-limit` consumes, and `mean_ms/median_ms/min_ms/max_ms` match its `--time-stat` choices. Error is collected once and shared; timing is collected per device.

It lands in the **same** `<cache-root>/<cache_id_base><suffix>/` dir `collect_cache.py` fills (suffix from the estimator's `get_cache_id_suffix`), so error + timing sit together.

## Notes

- Runs in one process (no per-config subprocess), **skips** configs that already have a `timing.json` (`--overwrite` to force), and continues past a failing config unless `--halt-on-error` — mirroring `collect_cache.py`'s ergonomics. Reuses its `load_estimators_list`/`materialize_model_configs` helpers.
- On a shared GPU, export `XLA_PYTHON_CLIENT_PREALLOCATE=false` so JAX doesn't grab 75% of the device (collides with other jobs → cuBLAS-init failures). For trustworthy numbers, run when the GPU is otherwise idle.

## Testing

- Unit tests cover the JAX-free parts: record schema (the four `select_ensemble` fields), cache-id-base resolution, and arg parsing.
- `ruff` (0.14.1) check + format, `pydoclint`, `basedpyright` all clean.
- Functionally smoked on an RTX 5090 over real `openpiv_jax` sweep configs: produces valid `timing.json` consumed by `select_ensemble --time-limit`.

Stacked on #62 (base `feat/generic-cache-collector-public`), since `select_ensemble.py` — which reads the `timing.json` this writes — lives there. Rebase to `dev` once #62 merges.

### DIFF
--- a/scripts/collect_timing.py
+++ b/scripts/collect_timing.py
@@ -1,0 +1,468 @@
+"""Collect a per-config ``timing.json`` for a sweep of model configs.
+
+Sibling of ``collect_cache.py``. Where that wrapper fills the per-image
+**error** cache (device-independent, reusable anywhere), this one measures the
+**device-dependent** inference time and writes ``timing.json`` into each
+config's cache dir -- the field ``select_ensemble.py --time-limit`` consumes.
+The two are deliberately separate: error is reproducible across machines and
+cached once; timing depends on the GPU/driver/load and is collected per device.
+
+For every model config it builds the estimator with ``make_estimator`` (exactly
+as ``src/main.py --mode eval`` does, so the timed forward is the one a consumer
+pays for), then times the JIT-compiled forward the way ``tests/test_dis_jax.py``
+speed tests do:
+
+- a random input pair of the dataset's image shape is made **device-resident**
+  (on GPU when one is available) *before* timing -- PIV/optical-flow compute
+  cost is data-independent, so no data loading is needed;
+- the forward is JIT-compiled and warmed (compile excluded from the result);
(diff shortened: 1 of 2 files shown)


antonioterpin (dismissed):

albonomi02 (dismissed):
LGTM

PR #71, opened 2026-06-09
docs(examples): demo cache + timing + ensemble selection together

fbanelli:
## What

Adds **`examples/14_cache_sweep_ensemble.py`** — an end-to-end, CPU-only, fully synthetic demo of the three estimator-agnostic sweep tools working together:

1. `scripts/collect_cache.py` — per-image **error** cache (device-independent)
2. `scripts/collect_timing.py` — per-config **`timing.json`** (device-dependent, `timeit` at bs=1, input pre-loaded on device)
3. `scripts/select_ensemble.py` — pick a `K`-subset minimizing per-image best error (optionally under a latency bound), and `--export-models` the chosen configs as a collect-ready `estimators_list`.

It builds a tiny synthpix dataset (no downloads) and three `dis_jax` configs, then runs collect → time → select, prints the cache layout, and exports the winners.

## Why — demonstrates the three are inter-compatible

Each candidate maps to one cache dir `<cache-root>/<cache_id>/` (`cache_id` = base + the estimator's `get_cache_id_suffix`). `collect_cache` writes the error parquet under `data/`, `collect_timing` writes `timing.json` alongside it, and `select_ensemble` reads both — so with a shared `--cache-root` + base id they land together automatically. The demo prints this explicitly:

```text
=== cache layout (error + timing co-located per config) ===
  demo-sweep_c5a09d90c: error=True timing=True
  demo-sweep_cc712e59d: error=True timing=True
  demo-sweep_cd0430e37: error=True timing=True
```

and the selector consumes the timing (`#1 ... t=0.599ms ...`), then exports a config list that feeds straight back into a full-set `collect_cache --estimators-list` run.

## Docs

`docs/examples/sweep-ensemble.md` (linked from `docs/examples/index.md`).

## Testing

- Runs clean end-to-end via `uv run python examples/14_cache_sweep_ensemble.py` (CPU, ~synthetic, no downloads).
- `ruff` (0.14.1) check + format, `pydoclint`, `basedpyright` all clean on the new example.

Stacked on #70 (base `feat/generic-timing-collector`) because the demo exercises `collect_timing.py` from that PR (which itself stacks on #62). Rebase down the stack as those merge.

### DIFF
--- a/docs/examples/index.md
+++ b/docs/examples/index.md
@@ -23,6 +23,11 @@ order most readers will want to follow.
   `examples/11_caching.py`. Cold/warm cache reuse with the DIS and RAFT
   demos. Treat as an advanced optimization topic after the core
   eval/training paths make sense.
+- [Sweep, time, and select an ensemble](sweep-ensemble.md) —
+  `examples/14_cache_sweep_ensemble.py`. Runs `collect_cache.py`,
+  `collect_timing.py`, and `select_ensemble.py` together over a small
+  synthetic sweep: error cache, per-config timing, and a `K`-subset
+  selection that exports a collect-ready config list.
 
 `src/main.py` is the CLI entrypoint these workflows invoke.
 
@@ -36,4 +41,5 @@ flow-eval
 supervised-training
 density-eval
 caching
+sweep-ensemble
 ```
--- a/docs/examples/sweep-ensemble.md
+++ b/docs/examples/sweep-ensemble.md
@@ -1,0 +1,61 @@
+# Sweep, time, and select an ensemble
+
+`examples/14_cache_sweep_ensemble.py` demonstrates the three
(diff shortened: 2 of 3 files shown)


antonioterpin (commented):
Need to run an agent to check that docs are not too duplicated also in docstrings of the scripts used here

albonomi02 (dismissed):
LGTM

antonioterpin (dismissed):
Ran the doc-duplication check from the open thread. The example's module docstring duplicated `docs/examples/sweep-ensemble.md` near-verbatim (the three-tool list + the inter-compatibility paragraph) — a drift hazard. Trimmed it to a one-line purpose plus a pointer in `af27af4`, keeping `sweep-ensemble.md` as the single source of truth. Where the `.md` restates the scripts' own docstrings it's just one-line framing, which is fine. LGTM.

PR #73, opened 2026-06-10
feat(raft): port RAFT256-PIV in JAX and PyTorch with shared converter

antonioterpin:
## What

Adds the **RAFT256-PIV** variant of RAFT alongside the existing RAFT32-PIV,
in both JAX (Flax) and PyTorch, mirroring exactly what we did for RAFT32
(PR #47). The PyTorch model is the reference for crediting and bit-exact
comparison; the JAX model is the production estimator.

Ported from the reference Code Ocean capsule
[7226151](https://codeocean.com/capsule/7226151/tree/v1) — Lagemann et al.,
*Deep recurrent optical flow learning for PIV data*, Nature Machine
Intelligence 2021 ([doi](https://doi.org/10.1038/s42256-021-00369-0)); built
on Teed & Deng, *RAFT*, ECCV 2020 ([arXiv](https://arxiv.org/abs/2003.12039)).

## How it differs from RAFT32 (and why the converter is reused)

RAFT256-PIV is identical to RAFT32-PIV except:
- the feature/context encoder downsamples the 256×256 window to 32×32
  (stride-2 in each of the three residual stages), and
- the final flow is recovered with the learned **convex upsampling** head.

Neither change alters parameter shapes or names, so the RAFT256 state dict —
and the resulting Flax parameter tree — is **byte-identical** to RAFT32's.
The existing `convert_raft_torch_state_dict` converter is reused unchanged.

## Reuse

- PyTorch: `ResidualBlock`, `BasicUpdateBlock`, `CorrBlock` and the coordinate
  helpers are imported from the RAFT32 modules; only the strided
  `BasicEncoder256` is new.
- JAX: `EncoderBlock` is parameterized with strides + explicit PyTorch-matching
  padding (defaults keep RAFT32 byte-identical); `UpdateBlock`,
  `build_corr_pyramid` and the scan structure are reused.
- Estimators subclass `RaftJaxEstimator` / `RaftTorchEstimator` via small
  overridable build hooks, reusing the full patchify/fold + training + caching
  pipeline.

## Bit-perfect verification

The port is mathematically exact. Against the **real RAFT256-PIV ProbClass1
checkpoint**, running the Flax model with float32 inner convs and
`jax_default_matmul_precision='highest'` reproduces the PyTorch output to
**max 5.7e-4** absolute error (0 missing/unexpected keys). The default
mixed-precision config (float16 inner convs, as in RAFT32) lands in the same
tolerance regime as RAFT32, amplified by the convex-upsampling softmax over
16 iterations.

## Tests

`tests/unit/nn/test_raft256.py`:
- `convex_upsample` matches an explicit-loop reference (and is float64-exact
  vs PyTorch `upsample_flow`);
- RAFT256 Flax parameter tree is structurally identical to RAFT32's
  (transitively validating the shared converter);
- slow PyTorch parity tests over the last and all intermediate iterations.

Existing RAFT32 + blocks tests stay green; the JAX/PyTorch estimators are wired
into the registry (`raft_jax_256`, `raft_torch_256`) with configs and API docs.


antonioterpin:
### Real-data EPE parity (piv-class1 test set, ProbClass1 checkpoint)

Ran the converted JAX `RaftEstimatorModel256` and the PyTorch `RAFT256`
side by side on the local `piv-class1` test split with the published
RAFT256-PIV ProbClass1 weights (256×256 windows, 16 iterations,
**default mixed-precision config** — float16 inner convs). 6 images per
category:

| category | n | EPE_torch | EPE_jax | max·\|j−t\| |
|---|---|---|---|---|
| cylinder | 6 | 0.0113 | 0.0113 | 0.0015 |
| backstep | 6 | 0.0151 | 0.0151 | 0.0037 |
| DNS_turbulence | 6 | 0.0512 | 0.0512 | 0.0042 |
| SQG | 6 | 0.1066 | 0.1066 | 0.0064 |
| JHTDB_channel | 6 | 0.1216 | 0.1216 | 0.0043 |
| **OVERALL** | **30** | **0.0612** | **0.0612** | **0.0064** |

Mean per-image `|EPE_jax − EPE_torch|` = **2e-5 px**. The EPE is identical
to displayed precision in every category; the max per-pixel flow
difference stays ≤ 0.0064 px. On real particle images the float16
divergence is far smaller than on synthetic uniform-noise inputs, which
produce ill-conditioned over-confident features.

src/flowgym/flow/raft/raft256_jax.py
```diff
@@ -0,0 +1,61 @@
...
+from typing import Any
+
+from flax import linen as nn
+
+from flowgym.flow.raft.raft_jax import RaftJaxEstimator
+from flowgym.nn.raft256_model import RaftEstimatorModel256
+
+
+class RaftJax256Estimator(RaftJaxEstimator):
+    """RAFT256 estimator: 1/8-resolution convex-upsampling RAFT variant."""
```
antonioterpin: When referring to raft, we should instead refer to raft32

antonioterpin: or better, keep raft but make clear default is raft32. so that raft is like the general scaffolding/structure

albonomi02: I agree, let's consider raft32 the default, and let's be specific everywhere

antonioterpin: Done in `70fb546`.

Went with **"keep `raft`, make the default explicit"**: unqualified `raft` now denotes the shared scaffolding/family (module, base estimator classes, NN blocks, PyTorch→Flax converter), and every *variant-specific* surface says RAFT32-PIV / RAFT256-PIV.

- `raft_jax` / `raft_torch` stay as the documented **default (= RAFT32-PIV)**, plus new explicit aliases `raft_jax_32` / `raft_torch_32` (and `RaftJax32Estimator` / `RaftTorch32Estimator`) so callers can name the variant unambiguously.
- New `raft_jax_32/` + `raft_torch_32/` config dirs; old `raft/` and `raft_jax/` kept.
- Docstrings + doc pages relabeled (titles, cross-refs).

Additive on purpose — the merged keys and the checkpoint `load_from` paths that embed `raft_jax/` keep working, so no migration. If you'd later rather steer people off the bare names toward the explicit `_32` ones, I can add a soft-deprecation note.

